PR #40, opened 2017-08-16
Add Travis CI

joshlory:


package.json
```diff
@@ -1,21 +1,21 @@
 {
-  "name": "phaser-grunt-boilerplate",
+  "name": "craft",
```
Hamms: do we want this to be `@code-dot-org/craft` to match our other npm-installed projects?

joshlory: Ah, good call.

PR #41, opened 2017-08-16
Basic unit test skeleton

joshlory:
I chose to use `tap` which is a minimal test runner for node (and will run in Travis CI without any extra dependencies).


joshlory:
I'm leaning towards no transpile step so the unit tests run as fast as possible.  I think we'll want to transpile and run in Headless Chrome for the integration tests, though.

PR #44, opened 2017-08-18
Dynamic redstone sprite determination

WilliamHarmonMS:


src/js/game/LevelMVC/LevelView.js
```diff
@@ -2015,6 +2025,102 @@ module.exports = class LevelView {
...
 
+  determineRedstoneSprite(x, y, blockType, levelData) {
+        let foundAbove = false;
+        let foundBelow = false;
+        let foundRight = false;
+        let foundLeft = false;
+        let belowIndex = (this.yToIndex(y + 1)) + x;
+        let aboveIndex = (this.yToIndex(y - 1)) + x;
+        let leftIndex = (this.yToIndex(y)) + x - 1;
+        let rightIndex = (this.yToIndex(y)) + x + 1;
```
joshlory: Nit: rest of this file uses two spaces for indent.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -2015,6 +2025,102 @@ module.exports = class LevelView {
...
+
+        let borderCount = 0;
+
+        //need to ensure these are in bounds
+        if (y === levelData.planeHeight) {
```
joshlory: Please use `levelModel.inBounds`.  Rather than referencing it through the controller, I think it would make more sense for this function to live in `LevelModel.js`.

joshlory: Hmm, you're right: it may indeed make more sense for the model to store "redstone_wire" everywhere, and the view to determine how to render it.  Let me think about this a little more.

joshlory:
I resolved conflicts from my changes in the base branch.  You can revert https://github.com/code-dot-org/craft/pull/44/commits/dc7604065f8c77079c91c3c140e65ab77701c227 if you want to handle the merge yourself.  Thanks!

WilliamHarmonMS:
@joshlory  PTAL :)

src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -94,7 +94,7 @@ module.exports = class LevelBlock {
       this.isTransparent = true;
     }
 
-    if (blockType === "redstone_dust") {
+    if (blockType.substring(0, 12) === "redstoneWire") {
```
joshlory: Nit: `blockType.startsWith("redstoneWire")`.

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -46,6 +46,15 @@ module.exports = class LevelModel {
...
     this.shadingPlane = [];
     this.actionPlane = new LevelPlane(this.initialLevelData.actionPlane, this.planeWidth, this.planeHeight, true);
+
+    for (let i = 0; i < this.actionPlane.length; ++i) {
+      if (this.actionPlane[i].blockType.substring(0,12) === "redstoneWire") {
+        let y = Math.floor(i / this.planeHeight);
+        let x = i - (y * this.planeHeight);
+        this.determineRedstoneSprite(x, y, this.actionPlane[i]);
+      }
+    }
```
joshlory: Can we implement this in `LevelPlane` now that it exists?  I _think_ it makes more sense to have this functionality on the `actionPlane` object (so it updates automatically when `actionPlane.setBlockAt` is called) but not 100% sure.

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -1116,6 +1196,106 @@ module.exports = class LevelModel {
     }
   }
 
+  determineRedstoneSprite(x, y) {
+        let foundAbove = false;
```
joshlory: Nit: inconsistent indentation.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -17,6 +17,7 @@ module.exports = class LevelView {
     this.fluffPlane = null;
     this.fowPlane = null;
     this.collectibleItems = [];
+    this.listOfFlowers = [];
```
joshlory: Not used?

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -1116,6 +1196,106 @@ module.exports = class LevelModel {
...
+        let leftIndex = (this.yToIndex(y)) + x - 1;
+        let rightIndex = (this.yToIndex(y)) + x + 1;
+
+        let borderCount = 0;
+
+        if (!this.inBounds(x, y)) {
+            // If we're looking out of bounds, just leave, you goofball.
+            return;
+        }
+        // If in bounds, we want to see if any redstone is around the index in question
```
joshlory: Can we re-use `getOrthogonalBlocks` here?  Once you merge latest `master` into this branch you should have access to `LevelPlane.prototype.getOrthogonalBlocks`.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -926,6 +940,50 @@ module.exports = class LevelView {
...
+    if (blockType.substring(0,substring.length) === substring) {
+      let blockIndex = (this.yToIndex(destroyPosition[1])) + destroyPosition[0];
+      let upIndex = blockIndex - this.controller.levelModel.actionPlane.height;
+      let downIndex = blockIndex + this.controller.levelModel.actionPlane.height;
+      let rightIndex = blockIndex + 1;
+      let leftIndex = blockIndex - 1;
+
+      let activeIndicies = [];
+
+      // Up index.
```
joshlory: Same feedback as above: Can we re-use `getOrthogonalBlocks` here?

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1650,6 +1738,11 @@ module.exports = class LevelView {
       sprite = null,
       frameList;
 
+    // Need to make sure the switch case will capture the right miniBlock for -all- redstoneWire
+    if (blockType.substring(0,12) === "redstoneWire") {
+      blockType = "redstoneWire";
```
joshlory: Is this right?  Now that we have `redstoneWireHorizontal` etc. I think we want that as the block type?

src/js/game/GameController.js
```diff
@@ -1341,8 +1341,10 @@ class GameController {
     if (this.levelModel.isBlockOfTypeOnPlane(forwardPosition, "lava", placementPlane)) {
       soundEffect = () => this.levelView.audioPlayer.play("fizz");
     }
+
```
joshlory: Extra?

src/js/game/LevelMVC/LevelView.js
```diff
@@ -128,9 +128,19 @@ module.exports = class LevelView {
...
+      "redstoneWireVertical": ["blocks", "redstone_dust_line_v_off", -13, 0],
+      "redstoneWireUpRight": ["blocks", "redstone_dust_corner_BottomLeft_off", -13, 0],
+      "redstoneWireUpLeft": ["blocks", "redstone_dust_corner_BottomRight_off", -13, 0],
+      "redstoneWireDownRight": ["blocks", "redstone_dust_corner_TopLeft_off", -13, 0],
+      "redstoneWireDownLeft": ["blocks", "redstone_dust_corner_TopRight_off", -13, 0],
+      "redstoneWireTUp": ["blocks", "redstone_dust_cross_up_off", -13, 0],
+      "redstoneWireTDown": ["blocks", "redstone_dust_cross_down_off", -13, 0],
+      "redstoneWireTLeft": ["blocks", "redstone_dust_cross_left_off", -13, 0],
+      "redstoneWireTRight": ["blocks", "redstone_dust_cross_right_off", -13, 0],
+      "redstoneWireCross": ["blocks", "redstone_dust_cross_off", -13, 0],
```
joshlory: Tiny nit: with the minecart tracks I've found it a lot easier to refer to them by north/south/east/west instead of up/down/right/left, since "up" as in "above a block" is different from "up" as in "north of a block".  Might be clearer to follow that pattern here?

WilliamHarmonMS: These I just took the JSON names as they were given to me by Joel, since that what he named his files. What do you mean about up vs above a block? I could understand if we were 3D, but since we're just a 2D array of blocks, I felt like "up" or "above" was still pretty clear? I can go through and change it if you like.

Even if we don't I can use north/south from now on if you want? Just let me know.

WilliamHarmonMS:
@joshlory I think we're ready for a PTAL :)

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -51,7 +51,8 @@ module.exports = class LevelModel {
       if (this.actionPlane[i].blockType.substring(0,12) === "redstoneWire") {
         let y = Math.floor(i / this.planeHeight);
         let x = i - (y * this.planeHeight);
-        this.determineRedstoneSprite(x, y, this.actionPlane[i]);
+        let position = [x,y];
+        this.actionPlane.determineRedstoneSprite(position, i);
```
joshlory: I think we want to trigger this whenever `LevelPlane.prototype.setBlockAt` is called.

joshlory:
Nit: `checkConnectionOnPlace` is singular but `checkConnectionsOnDestroy` is plural.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1126,7 +1183,7 @@ module.exports = class LevelView {
         sprite = null;
         if (!levelData.actionPlane[blockIndex].isEmpty) {
           blockType = levelData.actionPlane[blockIndex].blockType;
-          sprite = this.createBlock(this.actionPlane, x, y, blockType);
+          sprite = this.createBlock(this.actionPlane, x, y, blockType, levelData);
```
joshlory: Extra param isn't used?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -73,4 +73,153 @@ module.exports = class LevelPlane extends Array {
 
     return `rails${variant}`;
   }
+
+  checkConnectionOnPlace(substring, blockType, LevelModel) {
```
joshlory: If we're going this route, please pass in `levelModel` to the constructor.  The level plane instance will always be associated with the same model.

joshlory:
This PR is getting close.  I've put together a quick diff with the major changes I'm asking for here, to describe what I mean: https://github.com/code-dot-org/craft/compare/dynamic-redstone...dynamic-redstone-feedback.

I don't recommend applying that diff as-is.  It's worth going through to understand and accept/reject each suggestion.  Let me know if you have any questions!

src/js/game/LevelMVC/LevelView.js
joshlory: I think north/south/east/west is clearer for the reason I mentioned above, but certainly not blocking for this PR.

src/js/game/LevelMVC/LevelView.js
joshlory: Ah, I see — this is for dropping mini blocks on destroy.

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: Roger that. I pass in the LevelModel when the function is called, so it's using that argument, but I added it to the constructor as well.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1909,6 +1974,14 @@ module.exports = class LevelView {
...
         break;
 
+      case "redstoneWire":
+        atlas = this.blocks[blockType][0];
+        frame = this.blocks[blockType][1];
+        xOffset = this.blocks[blockType][2];
+        yOffset = this.blocks[blockType][3];
+        sprite = plane.create(xOffset + 40 * x, yOffset + plane.yOffset + 40 * y, atlas, frame);
+        break;
+
```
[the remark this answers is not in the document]
WilliamHarmonMS: Ah right, I had the special case there because of redstone charge propagation I think, but I'll reserve that addition for that PR.

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: I think it was in one approach I took, but it looks like I ended up going a different route and forgot to remove that. Thanks!

src/js/game/LevelMVC/LevelPlane.js
joshlory: Why both?  Once it's passed into the constructor, saving `this.levelModel = levelModel` will make it available as an instance variable.

WilliamHarmonMS: Yeah, I see what you mean. I didn't register that the reference on construction would be just that: a reference, rather than copying a snapshop of the model. I have this working with your suggestion

WilliamHarmonMS:
I believe we're ready for another PTAL @joshlory

joshlory:
I'm seeing a small issue where the redstone wire block doesn't appear until a half-second after the neighboring blocks connect.  Do you repro?

redstone-placement-delay

joshlory:
The other issue I'm seeing is that the "Destroy_Overlay.png" asset always assumes the item being destroyed is a full-height block.  It covers too much area when destroying a flat sprite like a redstone wire or minecart rail.  (This shouldn't block check in!)

WilliamHarmonMS:
PTAL 👍

src/js/game/GameController.js
```diff
@@ -1144,9 +1144,11 @@ class GameController {
               blockType = "planksSpruce";
               break;
           }
-
           this.levelView.playDestroyBlockAnimation(player.position, player.facing, destroyPosition, blockType, () => {
             commandQueueItem.succeeded();
+            this.levelModel.actionPlane.getOrthogonalPositions(destroyPosition).forEach(orthogonalPosition => {
+              this.levelView.refreshActionPlane(this.levelModel, [this.levelModel.actionPlane.coordinatesToIndex(orthogonalPosition)]);
+            });
```
joshlory: I think the `setBlockAt` function should be responsible for refreshing the view?

src/js/game/GameController.js
```diff
@@ -1341,13 +1343,18 @@ class GameController {
...
     this.levelView.playPlaceBlockInFrontAnimation(this.levelModel.player.position, this.levelModel.player.facing, forwardPosition, placementPlane, placedBlock.blockType, () => {
+
       this.levelModel.computeShadingPlane();
       this.levelModel.computeFowPlane();
       this.levelView.updateShadingPlane(this.levelModel.shadingPlane);
       this.levelView.updateFowPlane(this.levelModel.fowPlane);
       soundEffect();
+      this.levelModel.actionPlane.getOrthogonalPositions(forwardPosition).forEach(orthogonalPosition => {
+        this.levelView.refreshActionPlane(this.levelModel, [this.levelModel.actionPlane.coordinatesToIndex(orthogonalPosition)]);
+      });
```
joshlory: Same feedback as above.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -1,11 +1,12 @@
 const LevelBlock = require("./LevelBlock.js");
 
 module.exports = class LevelPlane extends Array {
-  constructor(planeData, width, height, isActionPlane = false) {
+  constructor(planeData, width, height, isActionPlane = false, LevelModel = null) {
```
joshlory: Nit: camelCase

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -36,13 +37,43 @@ module.exports = class LevelPlane extends Array {
   setBlockAt(position, block) {
     this[this.coordinatesToIndex(position)] = block;
 
+    if (block.isRedstone) {
+      this.determineRedstoneSprite(position);
+      this.getOrthogonalPositions(position).forEach(orthogonalPosition => {
+        const orthogonalBlock = this.getBlockAt(orthogonalPosition);
+        if (orthogonalBlock && orthogonalBlock.isRedstone) {
+          this.determineRedstoneSprite(orthogonalPosition);
+          //this.levelModel.controller.levelView.refreshActionPlane(this.levelModel, [this.coordinatesToIndex(orthogonalPosition)]);
```
joshlory: Commented out?  If we do the refresh in this method we shouldn't need to refresh from the Controller.  But this would only refresh the touching blocks, not the placed (or destroyed) block.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1677,6 +1693,8 @@ module.exports = class LevelView {
         frame = blockType;
         break;
     }
+    //blockType is used later in this function, so reset to original
+    blockType = saveType;
```
joshlory: Why is this needed?

src/js/game/GameController.js
```diff
@@ -1115,7 +1115,7 @@ class GameController {
     let player = this.levelModel.player;
     // if there is a destroyable block in front of the player
     if (this.levelModel.canDestroyBlockForward()) {
-      let block = this.levelModel.destroyBlockForward();
+      let block = this.levelModel.actionPlane.getBlockAt(this.levelModel.getMoveForwardPosition(player));
```
joshlory: 👍

src/js/game/GameController.js
```diff
@@ -1341,13 +1341,14 @@ class GameController {
...
-    const placedBlock = this.levelModel.placeBlockForward(blockType, placementPlane);
-    this.levelView.playPlaceBlockInFrontAnimation(this.levelModel.player.position, this.levelModel.player.facing, forwardPosition, placementPlane, placedBlock.blockType, () => {
+
+    this.levelView.playPlaceBlockInFrontAnimation(this.levelModel.player.position, this.levelModel.player.facing, forwardPosition, placementPlane, blockType, () => {
       this.levelModel.computeShadingPlane();
       this.levelModel.computeFowPlane();
       this.levelView.updateShadingPlane(this.levelModel.shadingPlane);
       this.levelView.updateFowPlane(this.levelModel.fowPlane);
       soundEffect();
+      this.levelModel.placeBlockForward(blockType, placementPlane);
```
joshlory: I'm seeing a small bug with `gameController.codeOrgAPI.placeInFront(null, 'grass', 'Player')`.  The newly-placed block doesn't have a shadow until another block is placed.

I think the fix is just to move this up above the call to `computeShadingPlane`.

WilliamHarmonMS: good catch! That's absolutely it. I didn't even notice the graphical discrepancy.

PR #45, opened 2017-08-18
Agent movement differentiated from player movement

WilliamHarmonMS:
This has been done for a while and is pretty independent of other fixes. Let it sit here for as long as you like.


joshlory:
Is this change missing an `Entities/Agent.js` file?

PR #52, opened 2017-08-22
Reduce variance in code coverage metrics due to random animation timing

joshlory:
Seeing some minor changes in code coverage when I wouldn't expect anything to change.  The lines are entity animations that run after delaying a random number of seconds.  Try stubbing `Math.random()` to return a consistent value to reduce variance.

### DIFF
--- a/test/integration/AdventurerTest.js
+++ b/test/integration/AdventurerTest.js
@@ -1,5 +1,8 @@
 const test = require("tape");
+const sinon = require("sinon");
 const GameController = require("../../src/js/game/GameController");
+
+sinon.stub(Math, "random").returns(0.5);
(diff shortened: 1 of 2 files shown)


test/integration/AdventurerTest.js
```diff
@@ -1,6 +1,9 @@
 const test = require("tape");
+const sinon = require("sinon");
 const GameController = require("../../src/js/game/GameController");
 
+sinon.stub(Math, "random").returns(0.5);
```
Bjvanminnen: 😆

joshlory: https://xkcd.com/221/

PR #54, opened 2017-08-23
Dynamic tracks

joshlory:
Basic track placement following the rules of https://minecraft.gamepedia.com/Rail#Placement.

This PR doesn't yet update surrounding tracks.  That behavior will depend on @WilliamHarmonMS's change to refresh the Action Plane.  For now only the placed track is mutated to the correct sprite.

screen shot 2017-08-22 at 5 29 48 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,9 @@
   "description": "Top-down 2D game engine for the Craft app type. See NOTICE for restrictions on use.",
   "main": "build/js/app.bundled.js",
   "scripts": {
-    "codecov": "./node_modules/.bin/codecov",
-    "test:unit": "./node_modules/.bin/istanbul cover ./test/unit/*.js --report lcovonly --dir ./coverage/unit",
-    "test:integration": "./node_modules/.bin/karma start",
+    "codecov": "codecov",
+    "test:unit": "istanbul cover tape ./test/unit/*.js --report lcovonly --dir ./coverage/unit",
+    "test:integration": "karma start",
     "test": "npm run lint && npm run test:unit && npm run test:integration",
     "lint": "eslint --ext .js,.jsx src/js"
   },
--- a/src/js/game/GameController.js
+++ b/src/js/game/GameController.js
@@ -1254,9 +1254,11 @@ class GameController {
       if (blockTypeAtPosition !== "") {
         this.levelModel.destroyBlock(blockIndex);
       }
-      if (this.levelModel.placeBlock(blockType)) {
+
+      const placedBlock = this.levelModel.placeBlock(blockType);
+      if (placedBlock) {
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -4,9 +4,9 @@
   "description": "Top-down 2D game engine for the Craft app type. See NOTICE for restrictions on use.",
   "main": "build/js/app.bundled.js",
   "scripts": {
-    "codecov": "./node_modules/.bin/codecov",
-    "test:unit": "./node_modules/.bin/istanbul cover ./test/unit/*.js --report lcovonly --dir ./coverage/unit",
-    "test:integration": "./node_modules/.bin/karma start",
+    "codecov": "codecov",
+    "test:unit": "istanbul cover tape ./test/unit/*.js --report lcovonly --dir ./coverage/unit",
```
joshlory: Added because `istanbul cover ./test/unit/*.js` was only running a single unit test file.

src/js/game/GameController.js
```diff
@@ -1254,9 +1254,11 @@ class GameController {
       if (blockTypeAtPosition !== "") {
         this.levelModel.destroyBlock(blockIndex);
       }
-      if (this.levelModel.placeBlock(blockType)) {
+
+      const placedBlock = this.levelModel.placeBlock(blockType);
```
joshlory: This is following the pattern in https://github.com/code-dot-org/craft/pull/44/files#diff-bc9f268bfb25698c75c4794fdf5c069fR1347.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -15,15 +15,62 @@ module.exports = class LevelPlane extends Array {
     }
   }
 
+  inBounds(position) {
```
joshlory: In the future we should probably update all `inBounds` calls to reference this method, and remove `GameController.prototype.inBounds`.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -15,15 +15,62 @@ module.exports = class LevelPlane extends Array {
...
+      (comparator(orthogonal.east) << 2) +
+      (comparator(orthogonal.west) << 3)
+    );
+  }
+
+  determineRailType(position) {
+    const mask = this.getOrthogonalMask(position, block => block && block.isRail);
+    const variant = [
+      "Vertical", "Vertical", "Vertical", "Vertical",
+      "Horizontal", "BottomLeft", "TopLeft", "TopLeft",
```
joshlory: The corner names in the sprite sheet refer to sides of a square, so `BottomLeft` would be "┗".  A bit confusing!

balderdash: Ohh I get it now.

PR #55, opened 2017-08-24
Action plane refresh

WilliamHarmonMS:
I think I took care of all the merge conflicts without borking anything. I'm at just over 11 hours today though, so if my tired eyes missed something, just mark the feedback. 

refreshActionPlane() is probably a bad name for something that just refreshes specific indicies. I can fix that...

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -1158,14 +1158,12 @@ module.exports = class LevelView {
       }
     }
   }
-  
+
   refreshActionPlane(levelData, indices) {
-      var sprite,
+    var sprite,
       blockType;
-      
-    let whatever = this.actionPlaneBlocks;
-    let whatever2 = this.controller.levelModel.actionPlane;
-    for(var index in indices) {
+
+    for (var index in indices) {
       if (indices[index] >= 0) {
         let workingIndex = indices[index];
 
@@ -1188,29 +1186,33 @@ module.exports = class LevelView {
         this.actionPlaneBlocks[workingIndex] = sprite;
       }
     }
-    
+
+    // Save all all non-block elements.
     let entityHolder = [];
-    for(var index in this.actionPlane.children) {
-      let workingKey = this.actionPlane.children[index].key;
-      if(workingKey !== "blocks" && workingKey !== "torch" && workingKey !== "door"){
-        entityHolder.push(this.actionPlane.children[index]);
+    for (var aPIndex in this.actionPlane.children) {
+      let workingKey = this.actionPlane.children[aPIndex].key;
+      if (workingKey !== "blocks" && workingKey !== "torch" && workingKey !== "door"){
+        entityHolder.push(this.actionPlane.children[aPIndex]);
       }
     }
-    let whatamI = this.controller.levelModel.groundDecorationPlane;
-    
+
+    // Empty the actionPlane.
     this.actionPlane.children = [];
-    
-    for(var index in this.actionPlaneBlocks) {
-      if(this.actionPlaneBlocks[index] !== null) {
-        this.actionPlane.children.push(this.actionPlaneBlocks[index]);
-      }
-    }    
-    
-    for(var index in entityHolder) {
-      if(entityHolder[index] !== null) {
-        this.actionPlane.children.push(entityHolder[index]);
+
+    // Re-push all actionPlaneBlocks into actionPlane.
+    for (var aPBIndex in this.actionPlaneBlocks) {
+      if (this.actionPlaneBlocks[aPBIndex] !== null) {
+        this.actionPlane.children.push(this.actionPlaneBlocks[aPBIndex]);
       }
     }
+
+    // Re-capture all those non-blocks we saved.
+    for (var eHIndex in entityHolder) {
+      if (entityHolder[eHIndex] !== null) {
+        this.actionPlane.children.push(entityHolder[eHIndex]);
+      }
+    }
+    // We did all that to make sure everything in actionPlane is numbered sequentially.
   }
 
   updateShadingPlane(shadingData) {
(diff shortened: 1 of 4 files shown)


WilliamHarmonMS:
I am already aware that travis-ci is failing because of indents and white space and stuff. Will fix it first thing tomorrow. Let me know if you see anything terrible about the logic though.

src/test-build-only/levels.js
```diff
@@ -76,6 +76,7 @@ window.demoLevels = {
       "", "", "", "", "", "", "", "", "", "",
       "tallGrass", "", "", "", "", "", "", "", "", ""],
 
+
```
joshlory: Not needed?

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1159,6 +1159,62 @@ module.exports = class LevelView {
     }
   }
 
+  refreshActionPlane(levelData, indices) {
```
joshlory: Would be great to have some test coverage of this method!

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1159,6 +1159,62 @@ module.exports = class LevelView {
...
+    let entityHolder = [];
+    for (var aPIndex in this.actionPlane.children) {
+      let workingKey = this.actionPlane.children[aPIndex].key;
+      if (workingKey !== "blocks" && workingKey !== "torch" && workingKey !== "door"){
+        entityHolder.push(this.actionPlane.children[aPIndex]);
+      }
+    }
+
+    // Empty the actionPlane.
+    this.actionPlane.children = [];
```
joshlory: Instead of emptying the entire `actionPlane`, can we selectively remove the sprites that correspond to the `indices` passed in?  I'm thinking something like this:

```js
indices.forEach(index => {
  const sprite = this.actionPlaneBlocks[index];
  this.actionPlane.remove(sprite);
  // ... create replacement sprite
});
```

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1159,6 +1159,62 @@ module.exports = class LevelView {
...
+
+    for (var index in indices) {
+      if (indices[index] >= 0) {
+        let workingIndex = indices[index];
+
+        sprite = this.actionPlaneBlocks[workingIndex];
+        this.actionPlane.remove(sprite);
+
+        let y = Math.floor(workingIndex / this.controller.levelModel.planeHeight);
+        let x = workingIndex - (y * this.controller.levelModel.planeHeight);
```
joshlory: Rather than converting index -> x, y can we pass in an array of positions?  It's easy to go from [x, y] -> index via `coordinatesToIndex`.

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: The problem I encountered with this is that actionPlane has elements numbered sequentially, so if you remove a sprite from the middle (say actionPlaneBlocks[index] corresponds to actionPlane[6] or something), then you'll get a gap in actionPlane.

example:
actionPlane = {0: <sprite>, 1: <sprite>, 3: <sprite>, ......};

what -that- does it kill phaser.js preUpdate/postUpdate/Update functions, because they are expecting linearly numbered indicies. It loops through sprites to draw with for (let i = 0; i < size; ++i). Thus when it hits '2' in the example above, it errors.

The dumping and repushing is so that everything in view.actionPlane remains numbered properly.

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: I could. I generally don't consider a bit of math a big deal. If it doesn't exist, it seems silly to not have an indexToCoordinates, provided we have a coordinatesToIndex. Would you like me to put that in?

src/js/game/LevelMVC/LevelView.js
joshlory: How were you deleting the key at 2?  If you use `this.actionPlane.remove(sprite);` it should fix up the indices for you.

joshlory:
This is quite a bit more complicated than https://github.com/code-dot-org/craft/compare/dynamic-redstone-next#diff-894dfbadc994764b5b510d470f495021R1192 and the benefit of this approach isn't clear.

WilliamHarmonMS:
@joshlory This should be good to go, unless I've forgotten something. It is admittedly much more concise :)

src/js/game/LevelMVC/LevelView.js
joshlory: Sounds good, let's extract an `indexToCoordinates` helper function.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1159,6 +1159,31 @@ module.exports = class LevelView {
     }
   }
 
+  refreshActionPlane(levelData, indices) {
+    var sprite,
+      blockType;
+
+    for (var index in indices) {
+      if (indices[index] >= 0) {
```
joshlory: If you use `for (let i = 0; i < indices.length; i++)` instead of `for (... in ...)` you shouldn't need this check.  In JS the `for ... in` operator will enumerate all enumerable properties, which includes things like [].length.

Another (good) option would be to use `indices.forEach(index => { ... })`.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1159,6 +1159,31 @@ module.exports = class LevelView {
...
+
+        sprite = this.actionPlaneBlocks[workingIndex];
+        this.actionPlane.remove(sprite);
+
+        let y = Math.floor(workingIndex / this.controller.levelModel.planeHeight);
+        let x = workingIndex - (y * this.controller.levelModel.planeHeight);
+
+        blockType = levelData.actionPlane[workingIndex].blockType;
+        if (blockType !== "") {
+          this.createActionPlaneBlock({0: x,1: y}, blockType);
```
joshlory: Tiny nit: missing space after comma.

joshlory:
What's the test coverage plan for this change?

WilliamHarmonMS:
@joshlory PTAL 👍

PR #56, opened 2017-08-24
Events test coverage

joshlory:
Add first integration test for "Designer" levels (where `isEventLevel` = true).


joshlory:
Same approach as PR https://github.com/code-dot-org/craft/pull/47, but refactored to have both Adventurer and Designer level tests.

src/js/game/API/CodeOrgAPI.js
```diff
@@ -65,6 +65,20 @@ module.exports.get = function (controller) {
...
     },
 
+    onEventTriggered: function (highlightCallback, type, eventType, callback) {
+      this.registerEventCallback(highlightCallback,
+        function (event) {
+          if (event.eventType !== eventType) {
+            return;
+          }
+          if (event.targetType !== type) {
+            return;
```
islemaster: Uh... `if (event.eventType !=== eventType || event.targetType !== type)`?

src/js/game/GameController.js
```diff
@@ -219,13 +219,15 @@ class GameController {
...
 
+      tween.timeScale = 1;
       tween.start();
       tween = this.levelView.addResettableTween().to({
       }, this.timeout, Phaser.Easing.Linear.None);
 
       tween.onComplete.add(() => {
         this.endLevel(this.timeoutResult(this.levelModel));
       });
+      tween.timeScale = 1;
```
islemaster: Why is this needed?

test/integration/DesignerTest.js
```diff
@@ -0,0 +1,30 @@
...
+
+test('Designer 1: Chicken Move (pass)', t => {
+  attempt('designer01', api => new Promise(resolve => {
+    api.onEventTriggered(null, 'chicken', 2, event => {
+      api.moveForward(null, event.targetIdentifier);
+      api.turnLeft(null, event.targetIdentifier);
+    });
+
+    api.startAttempt((success, levelModel) => {
+      debugger;
```
islemaster: Did you mean to leave this `debugger` statement in?  If not, can we set our linter to catch this?

joshlory: 😳  whoops.  It looks like eslint is only running on `/src/js` — I will fix!

src/js/game/GameController.js
joshlory: Under integration tests, [the default `timeScale` is increased so tests run faster](https://github.com/code-dot-org/code-dot-org/pull/17114).

This tween is used to measure the level timeout... so I need to run it at normal speed.  In theory if all tweens are sped up by an equal amount they should still run in order, but it wasn't consistently passing.

I think when everything is run 100x there's some non-deterministic behavior of the timers (two callbacks may run in the same time slice, instead of sequentially).

islemaster: I'd suggest a comment to that effect, since in isolation (especially in src rather than test) this seems odd and it's not clear why we're doing it.

joshlory:
@islemaster PTAL!

PR #59, opened 2017-08-29
Continue adding test coverage for Events levels

joshlory:
Follow up to PR https://github.com/code-dot-org/craft/pull/56.


joshlory:
Good call!  I'll add some failure cases (getting exploded by a creeper, getting killed by zombies) to the later levels.  For the early levels the timeout code path is pretty similar so I think it should be covered by `Designer 1: Chicken Move (fail)`.

PR #61, opened 2017-08-30
Reduce Codecov root path

joshlory:
See: https://docs.codecov.io/v4.3.6/docs/fixing-paths


joshlory:
Before:

screen shot 2017-08-29 at 10 06 27 pm

After:

screen shot 2017-08-30 at 9 09 27 am

PR #66, opened 2017-08-31
Charge propagation

WilliamHarmonMS:
charge propagating through redstone on run, place, destroy


WilliamHarmonMS:
I wanted to make redstonePropagation loop through indices returned through getOrthogonalPositions(), but something with the way js handles the recursion made it stop working? Perhaps this is an area for improvement.

joshlory:
Can you resolve conflicts and fix the Travis CI failure for this branch?  Thanks!

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -51,6 +53,7 @@ module.exports = class LevelPlane extends Array {
           }
         }
       });
+      this.getRedstone();
```
joshlory: I think we also need to call this when placing `railsRedstoneTorch`?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -200,4 +203,75 @@ module.exports = class LevelPlane extends Array {
...
+      if (this[i].blockType === "railsRedstoneTorch") {
+        let y = Math.floor(i / this.height);
+        let x = i - (y * this.height);
+        let pos = [x,y];
+        this.redstonePropagation(pos);
+      }
+    }
+
+    for (let obj in this.redstoneList) {
+      console.log("The block at index: " + this.levelModel.controller.levelView.indexToCoordinates(this.redstoneList[obj]) + " is a " + this[this.redstoneList[obj]].blockType + ". This block is NOT charged.");
```
joshlory: I wonder if it's better to store a list of positions, so we don't have to convert from index to position?

Ideally we'd use `position` everywhere possible, only converting to index when looking up values in the internal array.  No external consumers should need to know the implementation of the underlying data structure.

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: That is a great point... last time I heard we -weren't- placing torches? I don't think? Something was going to be static last time I spoke with the designers. I'll absolutely add the case because who cares if the designers don't use it, but I'll confirm it anyway

PR #67, opened 2017-08-31
redstone charge propagation

WilliamHarmonMS:
Re-submitting the charge propagation tracking because the other branch got in a bad state.


src/js/game/LevelMVC/LevelView.js
```diff
@@ -481,6 +481,12 @@ module.exports = class LevelView {
     return (this.yToIndex(coordinates[1])) + coordinates[0];
   }
 
+  indexToCoordinates(index) {
+    let y = Math.floor(index / this.controller.levelModel.actionPlane.height);
+    let x = index - (y * this.controller.levelModel.actionPlane.height);
+    return [x,y];
+  }
```
joshlory: Please move this method to LevelPlane.js

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -229,4 +232,83 @@ module.exports = class LevelPlane extends Array {
 
     return this[myIndex].blockType;
   }
+
+  getRedstone() {
+    this.redstoneList = [];
+    this.redstoneListON = [];
+    for (let i = 0; i < this.length; ++i) {
+      if (this[i].blockType.substring(0,12) === "redstoneWire") {
```
joshlory: Tiny style nit: missing space after comma.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -229,4 +232,83 @@ module.exports = class LevelPlane extends Array {
...
+        let position = [];
+        if (this.levelModel !== null) {
+          position = this.levelModel.controller.levelView.indexToCoordinates(i);
+        }
+        this.redstonePropagation(position);
+      }
+    }
+
+    for (let obj in this.redstoneList) {
+      if (this.levelModel !== null) {
```
joshlory: Once you move `coordinatesToIndex` I don't think this null check is needed anymore?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -229,4 +232,83 @@ module.exports = class LevelPlane extends Array {
...
+    for (let obj in this.redstoneListON) {
+      if (this.levelModel !== null) {
+        console.log("The block at index: " + this.redstoneListON[obj] + " is a " + this[this.levelModel.controller.levelView.coordinatesToIndex(this.redstoneListON[obj])].blockType + ". This block IS charged.");
+      }
+    }
+  }
+
+  redstonePropagation(position) {
+    let block = this[this.levelModel.yToIndex(position[1]) + position[0]];
+    //console.log("the position of " + block.blockType + " is " + "(" + position.x + ", " + position.y + "). This block is charged!");
```
joshlory: Not needed?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -229,4 +232,83 @@ module.exports = class LevelPlane extends Array {
...
+    }
+    for (let obj in this.redstoneListON) {
+      if (this.levelModel !== null) {
+        console.log("The block at index: " + this.redstoneListON[obj] + " is a " + this[this.levelModel.controller.levelView.coordinatesToIndex(this.redstoneListON[obj])].blockType + ". This block IS charged.");
+      }
+    }
+  }
+
+  redstonePropagation(position) {
+    let block = this[this.levelModel.yToIndex(position[1]) + position[0]];
```
joshlory: Please use `this.coordinatesToIndex`

src/js/game/LevelMVC/LevelPlane.js
joshlory: A few more occurrences below.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -229,4 +232,83 @@ module.exports = class LevelPlane extends Array {
...
+      let indexToRemove = this.redstoneList.indexOf(position);
+      this.redstoneList.splice(indexToRemove,1);
+      this.redstoneListON.push(position);
+    }
+
+    //below current pos check
+    let adjacentBlock = this[this.levelModel.yToIndex(position[1] + 1) + position[0]];
+    if (this.levelModel.inBounds(position[0], position[1] + 1)
+      && adjacentBlock.isPowered === false
+      && adjacentBlock.blockType.substring(0,12) === "redstoneWire") {
```
joshlory: Tiny style nit: prefer the `&&` on the line before

```js
if (this.levelModel.inBounds(position[0], position[1] + 1) &&
  adjacentBlock.isPowered === false &&
  adjacentBlock.blockType.substring(0,12) === "redstoneWire") {
    // ...
}
```

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -229,4 +232,83 @@ module.exports = class LevelPlane extends Array {
...
+      this.redstonePropagation([position[0] - 1,position[1]]);
+    }
+    //right of current pos check
+    adjacentBlock = this[this.levelModel.yToIndex(position[1]) + position[0] + 1];
+    if (this.levelModel.inBounds(position[0] + 1, position[1])
+      && adjacentBlock.isPowered === false
+      && adjacentBlock.blockType.substring(0,12) === "redstoneWire") {
+      adjacentBlock.isPowered = true;
+      this.redstonePropagation([position[0] + 1,position[1]]);
+    }
```
joshlory: Nit: This block of very similar code is repeated 4 times.  Can we extract a helper method?

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: I get that block because redstonePropagation is called on wire and batteries (torches or pressure plates) and I only do the list pushing and popping if it's wire, because the rest are just sort of "always charged"

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -229,4 +238,54 @@ module.exports = class LevelPlane extends Array {
...
+    if (block.blockType.substring(0, 12) === "redstoneWire") {
+      let indexToRemove = this.redstoneList.indexOf(position);
+      this.redstoneList.splice(indexToRemove,1);
+      this.redstoneListON.push(position);
+    }
+
+    this.blockPropagation([position[0], position[1] + 1]);
+    this.blockPropagation([position[0], position[1] - 1]);
+    this.blockPropagation([position[0] + 1, position[1]]);
+    this.blockPropagation([position[0] - 1, position[1]]);
```
joshlory: This looks like a great candidate for calling `getOrthogonalPositions`!

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -144,6 +147,12 @@ module.exports = class LevelPlane extends Array {
     }
   }
 
+  indexToCoordinates(index) {
+    let y = Math.floor(index / this.width);
+    let x = index - (y * this.width);
+    return [x,y];
```
joshlory: Tiny style nit: missing space after comma (1 more below).

joshlory:
Should two of the redstone dust piles be lit when the level loads?

image

joshlory:
I'm not seeing redstone wire get powered when I place a redstone torch next to the wire:

image

test/unit/LevelPlaneTest.js
```diff
@@ -235,8 +245,44 @@ test('rail connections: destroy block', t => {
...
+  expected.height = undefined;
+  expected.levelModel = null;
+  expected.redstoneList = [];
+  expected.redstoneListON = [];
+
+  //the setBlockAt() calls should have run getRedstone() so the ON list ought to have
+  //3 elements while the not-ON list ought to have only a single element.
+  t.deepEqual(plane.map(block => block.blockType), expected);
+
+  t.end();
```
joshlory: Please add a test case for placing & destroying redstone torch!

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -227,6 +229,77 @@ module.exports = class LevelPlane extends Array {
...
+    }
+
     return this[myIndex].blockType;
   }
+
+  getRedstone() {
+    this.redstoneList = [];
+    this.redstoneListON = [];
+    for (let i = 0; i < this.length; ++i) {
+      if (this[i].blockType.substring(0, 12) === "redstoneWire") {
```
joshlory: Tiny style nit: `this[i].blockType.startsWith("redstoneWire")`

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -227,6 +229,77 @@ module.exports = class LevelPlane extends Array {
...
+        this[i].isPowered = false;
+        let position = this.indexToCoordinates(i);
+        this.redstoneList.push(position);
+      }
+    }
+    for (let i = 0; i < this.length; ++i) {
```
joshlory: We do this iteration a number of times — let's extract a `getMatchingPositions` helper (or similar) that iterates through the rows and column and takes a callback argument to determine if a given block matches.  The return value would be an array of matching positions.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -227,6 +229,77 @@ module.exports = class LevelPlane extends Array {
...
+        let position = this.indexToCoordinates(i);
+        this.redstonePropagation(position);
+      }
+    }
+
+    for (let i = 0; i < this.redstoneList.length; ++i) {
+      this.determineRedstoneSprite(this.redstoneList[i]);
+    }
+    if (this.levelModel !== null) {
+      this.levelModel.controller.levelView.refreshActionPlane(this.redstoneList);
```
joshlory: Does this mean we potentially refresh the same location twice?  (Once here, and once on line 76.)

As I've said before, I think this has the potential to introduce a lot of hard-to-track down bugs.  I’d prefer for the LevelPlane to have free reign changing blocks however it sees fit, then there’s one method at the end that compares the current state to the state of the sprites and does a batch update of the LevelView.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -227,6 +229,77 @@ module.exports = class LevelPlane extends Array {
...
+
+    for (let i = 0; i < this.redstoneListON.length; ++i) {
+      this.determineRedstoneSprite(this.redstoneListON[i]);
+    }
+    if (this.levelModel !== null) {
+      this.levelModel.controller.levelView.refreshActionPlane(this.redstoneListON);
+    }
+  }
+
+  findPositionInArray(position, array) {
```
joshlory: Style nit: this is equivalent to

```js
array.findIndex(element => element[0] === position[0] && element[1] === position[1]);
```

src/js/game/LevelMVC/LevelPlane.js
joshlory: Actually, why are we doing string compare on the block type?  Shouldn't these LevelBlock instances already have `this[i].isRedstone === true`?

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: Ah, yes. That's true. 76 should only be refreshing stuff if it -wasn't- a redstone block.

I agree that something more systemic in terms of comparing all blockTypes and refreshing only those indices at the end that don't match "last tick" or what not would be a fine idea. But I didn't know if that seemed like a change a bit more independent of the redstone functionality itself (it would certainly effect redstone, but it would effect every other placement/destruction/change to a plane in general so it didn't feel like this was the place?)

joshlory: If we're not doing that cleanup now, at minimum can we have this function return the list of positions that should be refreshed?

This way we're still calling the view refresh in a single centralized place.

src/js/game/GameController.js
```diff
@@ -133,6 +133,7 @@ class GameController {
     this.onDayCallback = levelConfig.onDayCallback;
     this.onNightCallback = levelConfig.onNightCallback;
 
+    //this.levelModel.actionPlane.getRedstone();
```
joshlory: Not used?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -38,15 +40,33 @@ module.exports = class LevelPlane extends Array {
     }
   }
 
+  /**
+  * Determines whether the position in question is within the bounds of the plane
```
joshlory: Tiny style nit: add punctuation to comments.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -156,22 +193,30 @@ module.exports = class LevelPlane extends Array {
 
     // If in bounds, we want to see if any redstone is around the index in question
     // Below index
-    if (orthogonalBlocks.south.block !== undefined && orthogonalBlocks.south.block.blockType.startsWith("redstoneWire")) {
+    if (orthogonalBlocks.south.block !== undefined &&
+    (orthogonalBlocks.south.block.isRedstone ||
+    orthogonalBlocks.south.block.blockType === 'railsRedstoneTorch')) {
```
joshlory: Would be good to encode as much of this into the `LevelBlock` model as we can.  Maybe we want `affectsRedstone` / `affectedByRedstone` flags?

I don't want this conditional to get even more gnarly when we add pressure plates 😄.

WilliamHarmonMS: I was thinking about having a helper function that we could add conditionals to, and just call within those if statements? Rather than go and add different flags to a bunch of different blocks?

Either solution would be fine with me.

joshlory: Think of `LevelBlock` as an extensible type.  Right now we do `new LevelBlock("redstoneWire")` but there's no reason we couldn't change that to `new RedstoneWireBlock()` (where `RedstoneWireBlock extends LevelBlock`).

Following this pattern, we'll want to use polymorphism to track how each type behaves, vs. adding a helper function in `LevelPlane`.  Then we can ask a block about its behavior anywhere, without requiring the helper function.

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: That should be do-able. Let me play with some places that make sense in a way that won't disrupt behavior.

src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -13,6 +13,9 @@ module.exports = class LevelBlock {
     this.isEmissive = false;
     this.isTransparent = false;
     this.isRedstone = false;
+    this.isPowered = false;
+    this.isConnectedToRedstone = false;
+    this.isRedstoneBattery = false;
```
joshlory: Can you add a comment describing what these last two flags mean?

PR #71, opened 2017-09-01
Cleanup player offset magic numbers

joshlory:
The Agent character will have different default offsets than the player.  Extract a `positionToScreen` helper so the x and y offsets all live in one place.

There should be no functional difference from this change.

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -684,20 +684,7 @@ module.exports = class LevelView {
       tween = this.addResettableTween(this.player.sprite).to(
         this.positionToScreen(position, isOnBlock), 180, Phaser.Easing.Linear.None);
     } else {
-      animName = "jumpDown" + direction;
-      this.playScaledSpeed(this.player.sprite.animations, animName);
-      const start = this.positionToScreen(oldPosition);
-      const end = this.positionToScreen(position);
-      tween = this.addResettableTween(this.player.sprite).to({
-        x: [start.x, end.x, end.x],
-        y: [start.y, end.y - 50, end.y],
-      }, 300, Phaser.Easing.Linear.None).interpolation((v, k) => {
-        return Phaser.Math.bezierInterpolation(v, k);
-      });
-
-      tween.onComplete.add(() => {
-        this.audioPlayer.play("fall");
-      });
+      tween = this.playPlayerJumpDownVerticalAnimation(position, oldPosition, direction);
     }
 
     tween.onComplete.add(() => {
@@ -705,8 +692,6 @@ module.exports = class LevelView {
     });
 
     tween.start();
-
-    return tween;
   }
 
   playPlayerJumpDownVerticalAnimation(position, oldPosition, direction) {
@@ -724,6 +709,8 @@ module.exports = class LevelView {
       this.audioPlayer.play("fall");
     });
     tween.start();
+
+    return tween;
   }
 
   playPlaceBlockAnimation(position, facing, blockType, blockTypeAtPosition, completionHandler) {
@@ -1052,7 +1039,7 @@ module.exports = class LevelView {
     return {
       x: -18 + 40 * x,
       y: -32 + (isOnBlock ? -23 : 0) + 40 * y,
-    }
+    };
   }
 
   setPlayerPosition(x, y, isOnBlock) {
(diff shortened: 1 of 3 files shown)


src/js/game/LevelMVC/LevelView.js
```diff
@@ -1059,10 +1034,19 @@ module.exports = class LevelView {
     tween.start();
   }
 
+  positionToScreen(position, isOnBlock = false) {
```
Bjvanminnen: Worth a short function description/param tags?

joshlory:
The Codecov report for this change is a little misleading: we only hit these lines in integration tests, which don't actually validate what is rendered.

You can watch the integration tests (to manually verify) with `WATCH=1 npm run test:integration`. 
 The default step speed is set in [`RunLevel.js`](https://github.com/code-dot-org/craft/blob/fefb497249a99bfb3caf87d32da7775c76a84136/test/helpers/RunLevel.js#L21).

src/js/game/LevelMVC/LevelView.js
Bjvanminnen: Feels a little strange that we take [x,y] (an array) as input, but then output {x, y} (an object).

src/js/game/LevelMVC/LevelView.js
```diff
@@ -682,58 +677,40 @@ module.exports = class LevelView {
...
+    var animName = "jumpDown" + direction;
     this.playScaledSpeed(this.player.sprite.animations, animName);
-    var tween = this.addResettableTween(this.player.sprite).to({
-      x: [-18 + 40 * position[0], -18 + 40 * position[0], -18 + 40 * position[0]],
-      y: [-32 + 40 * position[1], -32 + 40 * position[1] - 50, -32 + 40 * position[1]]
+    const start = this.positionToScreen(oldPosition);
+    const end = this.positionToScreen(position);
+    const tween = this.addResettableTween(this.player.sprite).to({
+      x: [start.x, end.x, end.x],
+      y: [start.y, end.y - 50, end.y],
```
Bjvanminnen: To be the same as the old, shouldnt this be:
```
x: [start.x, start.x, start.x],
y: [end.y, end.y - 50, end.y]
```

src/js/game/LevelMVC/LevelView.js
```diff
@@ -682,58 +677,40 @@ module.exports = class LevelView {
...
     });
 
     tween.start();
-
-    return tween;
   }
 
-  playPlayerJumpDownVerticalAnimation(position, direction) {
-    var animName = "jumpDown" + this.getDirectionName(direction);
+  playPlayerJumpDownVerticalAnimation(position, oldPosition, direction) {
```
Bjvanminnen: again maybe worth a short function description/@param tags?

src/js/game/LevelMVC/LevelView.js
```diff
@@ -682,58 +677,40 @@ module.exports = class LevelView {
     //make sure to render high for when moving up after placing a block
     var zOrderYIndex = position[1] + (facing === FacingDirection.Up ? 1 : 0);
     this.player.sprite.sortOrder = this.yToIndex(zOrderYIndex) + 5;
-    oldPosition = [Math.trunc((this.player.sprite.position.x + 18) / 40), Math.ceil((this.player.sprite.position.y + 32) / 40)];
-    newPosVec = [position[0] - oldPosition[0], position[1] - oldPosition[1]];
-
-    //change offset for moving on top of blocks
-    if (isOnBlock) {
-      yOffset -= 22;
-    }
```
Bjvanminnen: was 22, now 23 with your new helper `positionToScreen`. do we care?

src/js/game/LevelMVC/LevelView.js
Bjvanminnen: Oh i see. position == oldPosition in many cases.

src/js/game/LevelMVC/LevelView.js
joshlory: Ah, good catch.  I think they're close enough?  `23px` is on the left, `22px` is on the right:

screen shot 2017-09-01 at 3 38 27 pm

FWIW I measured the height of the side of a block and it's actually `20px` 🤷‍♀️.

src/js/game/LevelMVC/LevelView.js
joshlory: Yeah, I think I prefer `{x, y}` everywhere.  There are a ton of places in the code that currently use `[x, y]`.

src/js/game/LevelMVC/LevelView.js
Bjvanminnen: I can tell the difference. Seems fine to me.

PR #73, opened 2017-09-01
Add "Agent" as a player entity

joshlory:
Right now the Agent is an instance of `Player` with a different name.  I'll need to do some cleanup to make Agent inherit from Player, and move a lot of the player sprite code out of `LevelView`.


src/js/game/LevelMVC/LevelView.js
```diff
@@ -1355,240 +1350,126 @@ module.exports = class LevelView {
...
+    if (playerName === 'Agent') {
+      this.generateAnimations(FacingDirection.Down, -1);
+      this.generateAnimations(FacingDirection.Right, 65);
+      this.generateAnimations(FacingDirection.Up, 130);
+      this.generateAnimations(FacingDirection.Left, 195);
+    } else {
+      this.generateAnimations(FacingDirection.Down, 0);
+      this.generateAnimations(FacingDirection.Right, 60);
+      this.generateAnimations(FacingDirection.Up, 120);
+      this.generateAnimations(FacingDirection.Left, 180);
```
Bjvanminnen: Are these offsets that we expect to eventually get from the player object? i.e. instead of having a check for player name do something like player.getOffsets()?

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1355,240 +1350,126 @@ module.exports = class LevelView {
...
+    this.player.sprite.animations.add('mineCart_up', Phaser.Animation.generateFrameNames("Minecart_", 9, 9, "", 2), frameRate, false);
+    this.player.sprite.animations.add('mineCart_turnleft_up', Phaser.Animation.generateFrameNames("Minecart_", 10, 10, "", 2), frameRate, false);
+    this.player.sprite.animations.add('mineCart_turnright_up', Phaser.Animation.generateFrameNames("Minecart_", 8, 8, "", 2), frameRate, false);
+  }
+
+  playerFrameName(n) {
+    return Phaser.Animation.generateFrameNames("Player_", n, n, "", 3);
+  }
+
+  generateAnimations(facing, offset) {
```
Bjvanminnen: would be great to at least have @param tags for these new methods

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1355,240 +1350,126 @@ module.exports = class LevelView {
...
+    const direction = this.getDirectionName(facing);
+    var frameList,
+      i,
+      singlePunch,
+      idleFrameRate = 10;
```
Bjvanminnen: nit s/var/let

src/js/game/LevelMVC/LevelView.js
joshlory: Yeah, I think `generateAnimations` will go on the `Player` class, then `Agent` will override its `preparePlayerSprite` to generate animations from the correct offsets.

I'm working with the artist to see if we can standardize the offsets between the different player sprite sheets but I think this will be fine for now.

PR #75, opened 2017-09-05
Remove unused `whilePathAhead` API

joshlory:
This API is referenced by Adventurer, but never called: https://github.com/code-dot-org/code-dot-org/search?q=whilePathAhead.

screen shot 2017-09-05 at 1 39 54 pm

screen shot 2017-09-05 at 1 54 47 pm


joshlory:
Hmm, Redshift may not have the full LevelSources data set.  That said, we don't have any blocks that generate `whilePathAhead` API calls: https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/craft/simple/blocks.js

joshlory:
I'm tempted to remove this for now.  It has no test coverage, and we can add it back (with appropriate tests) if we decide we need it.

joshlory:
I still think we should remove this, but I don't feel confidant about merging this PR before HoC 😄.

PR #77, opened 2017-09-05
Pressure plates new

WilliamHarmonMS:
Am aware that moveDirection() in the controller is not the optimal place for the pressurePlate swapping but that was the best place I found to line it up with the walking animations. We can go over and find a better place for it.


joshlory:
Missing assets?

WilliamHarmonMS:
Which assets? Pressure Plates have been in the blocks.png for a while I believe. The modification to the blocks array in LevelView should have established their sprites? Are you seeing something display inappropriately?

joshlory:
Ah sorry, you're right!

src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -111,6 +111,16 @@ module.exports = class LevelBlock {
...
     }
+
+    if (blockType.startsWith("pressurePlate")) {
+      this.isEntity = true;
+      this.isWalkable = true;
+      this.isUsable = true;
+      this.isDestroyable = false;
+      this.isTransparent = true;
+      this.isConnectedToRedstone = true;
+      this.isRedstoneBattery = blockType === 'pressurePlateUp' ? false : true;
```
joshlory: Nit: can simplify this to `blockType !== 'pressurePlateUp'` or `blockType === 'pressurePlateDown'`.

PR #78, opened 2017-09-06
Iron door implementation

WilliamHarmonMS:
Iron doors should work fully now.
Manual tests:
door opens/closes when stepping on/off pressure plates and when placing/destroying redstoneWire


src/js/game/LevelMVC/LevelView.js
```diff
@@ -1903,6 +1904,31 @@ module.exports = class LevelView {
...
+
+        frameList = [];
+        animationFramesIron = Phaser.Animation.generateFrameNames("DoorIron", 3, 0, "", 1);
+        for (let j = 0; j < 5; ++j) {
+          frameList.push("DoorIron3");
+        }
+        frameList = frameList.concat(animationFramesIron);
+        sprite.animations.add("close", frameList, 5, false);
+        break;
+
```
joshlory: Can we de-dupe this code with `case "door"` above?

joshlory:
If I destroy or place a block next to the door I'm seeing the door change state to "closed" when it should still be open:

image

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -311,10 +311,39 @@ module.exports = class LevelPlane extends Array {
       posToRefresh.push(this.redstoneListON[i]);
     }
 
+    //once we're done updating redstoneWire states, check to see if doors should open/close
```
joshlory: Tiny style nit: space after `//`, punctuation at end of comment.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -311,10 +311,39 @@ module.exports = class LevelPlane extends Array {
...
+    let player = this.levelModel.player;
+    this.levelModel.controller.levelView.setSelectionIndicatorPosition(this.indexToCoordinates(index)[0], this.indexToCoordinates(index)[1]);
+    this.levelModel.controller.audioPlayer.play("doorOpen");
+    // if it's not walable, then open otherwise, close
+    const canOpen = !this[index].isWalkable;
+    this.levelModel.controller.levelView.playDoorAnimation(this.indexToCoordinates(index), canOpen, () => {
+      this[index].isWalkable = !this[index].isWalkable;
+      this.levelModel.controller.levelView.playIdleAnimation(player.position, player.facing, player.isOnBlock);
+      this.levelModel.controller.levelView.setSelectionIndicatorPosition(player.position[0], player.position[1]);
+    });
```
joshlory: I have some concerns about make all these updates from the `LevelPlane` model.  Let's talk through the best place for the animation and selection indicator changes.

I don't have this fully thought out, but my instinct is that we want to keep the model as decoupled from the controller as we can.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -359,4 +388,18 @@ module.exports = class LevelPlane extends Array {
     }
   }
 
+  /**
+  * Checking power state for objects that are powered by redstone.
+  */
+  powerCheck(position) {
+    let amIPowered = false;
+    this.getOrthogonalPositions(position).forEach(orthogonalPosition => {
```
joshlory: Nit: you can use [`Array.prototype.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) here, which will also bail out early if the callback returns true.

joshlory:
In a few cases I'm seeing the door not fully open or close, but I can't reliably repro:

image

image

joshlory:
Should placing a pressure plate directly next to a door (without connecting redstone wire) cause the door to open when pressed?

image

src/js/game/LevelMVC/LevelPlane.js
joshlory: I chatted with @breville a bit about this.  Can we get to a place where we only pass the `LevelView` reference to the `LevelPlane` constructor?

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: The duping was mainly because all the sprite frame references are different. Door0 vs DoorIron0. 

I could take that block out into a helper function and pass in 'Door' vs 'IronDoor' and evaluate it before the switch statement? Let me know what you think the best way to go about that is.

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: Completely agree. animateDoor() was basically grabbed wholesale from the GameController for the case of use() on a standard door, and I'm just applying the same calls to IronDoor if it is powered.

In a sense, the animation of the door is akin to "the state of the door". For instance, if instead of an open animation, we just had an OpenDoor object and ClosedDoor object, I feel confident that LevelPlane would be where we managed the swapping (just like redstoneWire powered/unpowered). So it makes some sense for ActionPlane object states is managed by the plane, but... It's not just two objects, it is an animation, so that does make things a little different.

Totally open to moving this around if we think there is a better place.

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: I was looking at the stackoverflow bits about bailing out of loops in JS (basically just having a return statement within it to get out of the function entirely) but it wasn't working. Chances are that's just my being used to C++ not JS, and there being a built in way to achieve that doesn't surprise me. 👍 

If we want to make that function more streamlined I'm fine with it, but it does seem to have the desired functionality in its current state.

src/js/game/LevelMVC/LevelView.js
joshlory: I'm thinking something along the lines of:

```js
const buildDoor = (levelView, type) => {
  atlas = levelView.blocks[blockType][0];
  frame = levelView.blocks[blockType][1];
  xOffset = levelView.blocks[blockType][2];
  yOffset = levelView.blocks[blockType][3];
  sprite = plane.create(xOffset + 40 * x, yOffset + plane.yOffset + 40 * y, atlas, frame);

  frameList = [];
  var animationFramesIron = Phaser.Animation.generateFrameNames(type, 0, 3, "", 1);
  for (let j = 0; j < 5; ++j) {
    frameList.push(`${type}0`);
  }
  frameList = frameList.concat(animationFramesIron);

  sprite.animations.add("open", frameList, 5, false);

  frameList = [];
  animationFramesIron = Phaser.Animation.generateFrameNames(type, 3, 0, "", 1);
  for (let j = 0; j < 5; ++j) {
    frameList.push(`${type}3`);
  }
  frameList = frameList.concat(animationFramesIron);
  sprite.animations.add("close", frameList, 5, false);

  return sprite;
};

// ...

case "door":
  sprite = buildDoor(this, "Door");
  break;

case "doorIron":
  sprite = buildDoor(this, "IronDoor");
  break;
```

src/js/game/LevelMVC/LevelPlane.js
joshlory: I prefer this for readability:

```js
powerCheck(position) {
  return this.getOrthogonalPositions(position).some(orthogonalPosition => (
    this[this.coordinatesToIndex(orthogonalPosition)].isRedstone &&
    this[this.coordinatesToIndex(orthogonalPosition)].isPowered)
  );
}
```

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: Roger that. 👍

WilliamHarmonMS:
Door refresh when destroying a block next to it is fixed, as is the pressure-plate-right-next-to-door issue.

That second one I'm SUPER not in love with, because a ton of problems with either the player ending up on a closed door index (because it closed after he left the plate) or issues where if you walked from plate to door back to plate, this was significantly stickier than the refresh. I had to add a number of arguments to functions that didn't otherwise need them, which I hate doing, but in order to get position data about where the player was/is/moving to that's what ended up happening.

In short the flow is:
Controller move() ~> if the player is leaving or entering a pressure plate, then:
LevelPlane placeAt() ~> see if he's coming from a door, or entering a door, index. If so:
LevelPlane getRedstone() ~> ignore animating that door if he's on it or re-entering the same pressure plate that powered it, or animating it if he's moving to an empty, non-plate index.

garbage, but it seems to work.

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: Curious non-important question: `${type}0` vs type + "0"
Strings in js have the + operator function as a concatenation correct? Is there a reason for 1 vs the other? Or is it just readability / style?

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: Somehow did not see your second comment. We can absolutely pass the LevelView into the plane at construction. That should be easy peezy and work just fine. Is that something we want to refactor for this PR? Or just clean up in future polishing PRs?

WilliamHarmonMS:
Chagned door animation to live in levelView. One question is that npm test is passing all the tests, and the linter isn't throwing any errors, but I'm still seeing Travis-ci be grumpy about something. So not sure what that's about?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -284,7 +317,7 @@ module.exports = class LevelPlane extends Array {
   * Important note: This is what kicks off redstone charge propagation and is called
   * on place/destroy/run/load.... wherever updating charge is important.
   */
-  getRedstone() {
+  getRedstone(OnDoor = false, DoorPosition = null) {
```
Hamms: nit: variable names should be camelCase, not PascalCase

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1979,4 +1997,22 @@ module.exports = class LevelView {
     this.resettableTweens.push(tween);
     return tween;
   }
+
+  /**
+  * Animate Door and set the status
+  */
+  animateDoor(index, Open) {
```
Hamms: nit: again, camelCase, please

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1979,4 +1997,22 @@ module.exports = class LevelView {
...
+    this.controller.levelModel.actionPlane[index].isOpen = Open;
+    let player = this.controller.levelModel.player;
+    this.setSelectionIndicatorPosition(this.controller.levelModel.actionPlane.indexToCoordinates(index)[0], this.controller.levelModel.actionPlane.indexToCoordinates(index)[1]);
+    this.controller.audioPlayer.play("doorOpen");
+    // If it's not walable, then open otherwise, close.
+    const canOpen = !this.controller.levelModel.actionPlane[index].isWalkable;
```
Hamms: I'm a bit confused by this. We're setting the door's open state based on the argument being passed here, but then deciding which animation to play based on its walkability? Seems like we're putting ourselves at risk of these two concepts getting out of sync. What am I missing here?

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1177,6 +1178,13 @@ module.exports = class LevelView {
     positions.forEach(position => {
       if (position) {
         const newBlock = this.controller.levelModel.actionPlane.getBlockAt(position);
+
+        //we don't want to refresh doors. They're not destroyable, and refreshing
+        //will lead to bad animation states
+        if (newBlock.blockType.startsWith("door")) {
```
Hamms: I'd love to have a less fragile way of determining whether or not something is a door, even if that just means putting this exact same check into a `newBlock.isDoor()` method

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1747,6 +1755,33 @@ module.exports = class LevelView {
...
 
+    const buildDoor = (levelView, type) => {
+      atlas = this.blocks[blockType][0];
+      frame = this.blocks[blockType][1];
+      xOffset = this.blocks[blockType][2];
+      yOffset = this.blocks[blockType][3];
+      sprite = plane.create(xOffset + 40 * x, yOffset + plane.yOffset + 40 * y, atlas, frame);
+
+      frameList = [];
+      var animationFramesIron = Phaser.Animation.generateFrameNames(type, 0, 3, "", 1);
```
Hamms: nit: this (and all variables declared above) should be `consts`. I'd also slightly prefer to see the framelists kept separate for clarity as something like `openFrameList` and `closeFrameList` (and similarly for animationFramesIron), but that's probably not strictly necessary.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -311,9 +344,41 @@ module.exports = class LevelPlane extends Array {
...
   }
 
+  positionEquivalence(lhs, rhs) {
+    if (lhs[0] === rhs[0]) {
+      if (lhs[1] === rhs[1]) {
+        return true;
+      }
+    }
+    return false;
+  }
```
Hamms: is this used anywhere?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -311,9 +344,41 @@ module.exports = class LevelPlane extends Array {
       posToRefresh.push(this.redstoneListON[i]);
     }
 
+    // Once we're done updating redstoneWire states, check to see if doors should open/close.
+    // Do things a bit different
+    if (OnDoor) {
+      this.findDoorToAnimate(this.coordinatesToIndex(DoorPosition));
+    } else {
+      this.findDoorToAnimate(-1);
+    }
```
Hamms: This definitely seems like a weird place to have this logic; any reason it shouldn't just live in the `setBlockAt` method above?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -311,9 +344,41 @@ module.exports = class LevelPlane extends Array {
...
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
+  * Find a door to animate
+  */
+  findDoorToAnimate(notOffendingIndex) {
```
Hamms: nit; I'd much prefer this method be called something that makes it clear that it is not only finding the door but also triggering the animation; `findAndAnimateDoor` or something

src/js/game/GameController.js
```diff
@@ -728,8 +728,11 @@ class GameController {
     if (isMovingOffOf) {
       let position = this.levelModel.player.position;
       let block = new LevelBlock('pressurePlateUp');
-      this.levelModel.actionPlane.setBlockAt(position, block);
+      this.levelModel.actionPlane.setBlockAt(position, block, false, direction);
     }
+    let OutOfDoor = this.levelModel.isEntityOnBlocktype("Player", "doorIron");
```
Hamms: nit: camelCase

src/js/game/LevelMVC/LevelView.js
Hamms: oh, scratch the note about consts; I see now that these are declared above

WilliamHarmonMS: That code was simply copy pasted from how animations were handled in previous years. That can be refactored I suppose if we want if style or guidelines have changed? I might personally put that sort of refactoring in a future polish oriented PR? Let me know what your thoughts are on that.

As for all the variables being const (You're talking about the ones at the top of the function, like xOffset, atlas, and so on?) would that not cause problems with the fact that we're assigning values to them in several cases throughout the function? Is there some aspect of const variables in js that I'm unaware of? I assumed const would make them read-only

src/js/game/LevelMVC/LevelView.js

src/js/game/GameController.js
WilliamHarmonMS: 👍

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: That's fair. Checking strings is not my favorite method of verifying something. However, this whole code base has, what I'd comfortably guess as -hundreds-, of blockType checks to dictate various behavior. I can easily extrapolate that out, just a heads up though.

Josh and I have spoken quite a bit about how much this code base could benefit from some heavy house-keeping. Just probably best to tackle the systemic stuff while we're not working on an active release. 😄

src/js/game/LevelMVC/LevelView.js
Hamms: Putting that in a polish-oriented PR is probably fine; like I said, it's a preference for the sake of future clarity (so that future copy-pasters get better code 😄) rather than a requirement.

And you're completely correct about the `consts`; I didn't realize I wasn't looking at the whole method. These are fine as-is

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: Good catch! It was, but I refactored the door logic with the the pressure-plate-right-next-to-door scenario, and this was no longer needed 👍

src/js/game/LevelMVC/LevelView.js
Hamms: Yeah, I've been seeing a lot of it. I am very much in favor of the idea of trying to make small housekeeping improvements as we go, since it's so much harder to schedule that kind of work after the product is finished.

That being said, consider this comment just a suggestion.

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: It was only in there because the state of the door will only ever change when redstone is changing. Meaning that if getRedstone isn't called (so redstone circuits aren't changing), the doors don't even need to do work.

If anything getRedstone() is not a great function. It's more like... "Anything pertaining to redstone needs to re-evaluate its life because something might have changed".

It can likely be taken into placeBlockAt() without losing any functionality if that's more desirable. 👍

Hamms: Ahh, that makes sense. Presumably we want the doors to change when switches are flipped or when redstone is destroyed as well, right? Can I assume that a switch flip does not go through `placeBlockAt` but does go through `getRedstone`?

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: Much like the other animation logic above, this was taken wholesale from the GameController and the way that door handling was done there. isOpen was added for door specific logic with pressure-plate stuff. I would wager that we can replace the walkable check with the open check. I didn't alter too much of that so much have just missed the check. 👍

test/unit/LevelPlaneTest.js
```diff
@@ -433,3 +433,58 @@ test('torch charge: place block', t => {
...
+    true,false,false,
+  ];
+
+  expected.width = undefined;
+  expected.height = undefined;
+  expected.levelModel = null;
+  expected.redstoneList = [];
+  expected.redstoneListON = [];
+
+  t.deepEqual(plane.map(block => block.isOpen), expected);
```
joshlory: I think it's clearer to write this as:
```js
t.ok(plane.getBlockAt([1, 0]).isOpen);
t.ok(plane.getBlockAt([1, 2]).isOpen);
```

test/unit/LevelPlaneTest.js
```diff
@@ -433,3 +433,58 @@ test('torch charge: place block', t => {
...
+    false,false,false,
+  ];
+
+  expected.width = undefined;
+  expected.height = undefined;
+  expected.levelModel = null;
+  expected.redstoneList = [];
+  expected.redstoneListON = [];
+
+  t.deepEqual(plane.map(block => block.isOpen), expected);
```
joshlory: I think it's clearer to write this as:
```js
t.ok(plane.getBlockAt([1, 0]).isOpen);
t.false(plane.getBlockAt([1, 2]).isOpen);
```

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -373,4 +434,16 @@ module.exports = class LevelPlane extends Array {
     }
   }
 
+  /**
+  * Checking power state for objects that are powered by redstone.
+  */
+  powerCheck(position) {
+    return this.getOrthogonalPositions(position).some(orthogonalPosition => {
+      if (this[this.coordinatesToIndex(orthogonalPosition)]) {
+        const block = this[this.coordinatesToIndex(orthogonalPosition)];
```
joshlory: Tiny style nit: can we combine these two lookups?
```js
const block = this[this.coordinatesToIndex(orthogonalPosition)];
if (block) ...
```
Or
```js
const block = this[this.coordinatesToIndex(orthogonalPosition)];
return block && ...
```

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -329,6 +368,28 @@ module.exports = class LevelPlane extends Array {
   }
 
   /**
+  * Find a door to animate
```
joshlory: Nit: can you expand this JSDoc description?

test/unit/LevelPlaneTest.js
```diff
@@ -433,3 +433,61 @@ test('torch charge: place block', t => {
...
+    false,true,false,
+    false,false,false,
+    true,false,false,
+  ];
+
+  expected.width = undefined;
+  expected.height = undefined;
+  expected.levelModel = null;
+  expected.redstoneList = [];
+  expected.redstoneListON = [];
```
joshlory: `expected` isn't used any longer?  (1 more below.)

joshlory:
Nice, this is looking really good!  I'm still seeing one issue where the player can stand on a closed door in some cases:

iron-door

joshlory:
Hmm, I'm also having trouble placing Iron Doors?  (I know this isn't a core scenario, but it worked before — I'm curious what changed.)

joshlory:
Still seeing the door animation get stuck sometimes.  Were you able to track down this issue?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -87,12 +87,51 @@ module.exports = class LevelPlane extends Array {
   * Changes the block at a desired position to the desired block.
   * Important note: This is the cornerstone of block placing/destroying.
   */
-  setBlockAt(position, block) {
+  setBlockAt(position, block, direction = null) {
```
joshlory: Can we match the signature of `getBlockAt(position, offsetX = 0, offsetY = 0)` here?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -87,12 +87,51 @@ module.exports = class LevelPlane extends Array {
...
+    }
+
+    let positionInQuestion = [0,0];
+    // This will either be the pos the player is leaving or entering, depending on situation
+    if (this.levelModel) {
+      positionInQuestion = [this.levelModel.player.position[0] + offset[0], this.levelModel.player.position[1] + offset[1]];
+    }
+    let wasOnADoor = false;
+    // If the questionable position was a door, we want to do a few things differently.
+    if (this[this.coordinatesToIndex(positionInQuestion)].blockType === "doorIron") {
```
joshlory: Can this be `this.getBlockAt(positionInQuestion)`?  Let's avoid working in indices as much as possible.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -329,6 +368,28 @@ module.exports = class LevelPlane extends Array {
   }
 
   /**
+  * Find all iron doors in a level and evaluate if they need to be animated based on state
+  */
+  findDoorToAnimate(notOffendingIndex) {
```
joshlory: Prefer this to take a `position` argument instead of an index.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -329,6 +368,28 @@ module.exports = class LevelPlane extends Array {
...
+    for (let i = 0; i < this.length; ++i) {
```
joshlory: Let's iterate through the positions so we're not directly dealing with indices until we need to actually look up items in the array. See my feedback here: https://github.com/code-dot-org/craft/pull/67/files#r137057705.

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: Ha, absolutely. While the gears were turning, I forgot we had a helper for that. 😂

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: Do you mean like, 
for(i = 0....)
 for(j = 0...)
  if(this.getblockat([i,j]).blockType .... )

Or is there some better way to go through all the positions in the plane? I just did it that way because at the end of the day, the array is 1D so it seemed an easy way to just cover every index.

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: This is... do-able, yeah. The problem is "direction" as it exists in the controller (which again, this is all to do with that -dumb- plate-next-to-door case) is literally just 0,1,2, or 3. It's simply an int, that goes clockwise (so like if you're going up, direction == 0. Down and direction == 2).

I could, in the controller, evaluate the direction, case out the offset, and pass that in, but it seemed a bit weird to do that in the controller? I guess it could make sense to do it there, since that's really the only time direction matters, which makes doing the evaluation in setBlockAt() sort of weird...er

But turning a direction of 3 into an offset of [-1,0] would probably -have- to exist in one of the 2 functions.

joshlory: Ah, I see — in that case I think that's reasonable for now.

PR #91, opened 2017-09-08
Make it possible to place unpowered and powered rails

joshlory:
Placement rules are much simpler than normal rails. Unpowered rails never curve. If possible, they'll form an East-West connection, otherwise they'll remain North-South.

image

image


joshlory:
The same test case as above, in-game:

screen shot 2017-09-07 at 5 41 08 pm 2

PR #92, opened 2017-09-08
Use normal placement rules for Adventurer 13 rails

joshlory:
Removes the special `railMap` for Adventurer level 13 and just uses normal rail placement rules (now that those are implemented).  This results in two minor changes:
 - The first tracked placed will be vertical, then snap to horizontal once the second track is placed.
 - The corner track will be horizontal when placed, then snap to a curved track when the connecting track is placed.

### Before:

before

### After:

after


joshlory:
### Before:

before

### After:

after

src/js/game/GameController.js
```diff
@@ -1265,12 +1255,14 @@ class GameController {
...
         this.levelView.playPlaceBlockAnimation(this.levelModel.player.position, this.levelModel.player.facing, blockType, blockTypeAtPosition, () => {
-          let force = false;
           if (this.checkMinecartLevelEndAnimation() && blockType === "rail") {
-            blockType = this.checkRailBlock(blockType);
-            force = true;
+            if (this.levelModel.player.position[1] < 7) {
+              blockType = "railsUnpoweredVertical";
+            } else {
+              blockType = "rails";
+            }
```
Hamms: 👍 much cleaner

Would be nice to see a comment explaining the motivation for this oneoff logic

PR #93, opened 2017-09-09
Minecart camera pan

joshlory:
Request from Joel, and follow up to PR https://github.com/code-dot-org/craft/pull/92.

This is the level before the 20x20 free-play world.  Include a bit of horizontal camera pan during the minecart animation, so it's clear the camera can move in the next level.

adventurer-13-pan-actual-speed


src/js/game/LevelMVC/LevelView.js
```diff
@@ -546,6 +546,8 @@ module.exports = class LevelView {
     position = [3, 2];
 
     animation = this.playLevelEndAnimation(position, facing, isOnBlock, completionHandler, false);
+    this.game.world.setBounds(0, 0, this.controller.levelModel.planeWidth * 40, this.controller.levelModel.planeHeight * 40);
```
Hamms: should these bounds be set elsewhere? What happens if the player tries to move off the edge of the screen before and after these are set?

joshlory: Good catch — right now they can move off screen during the puzzle.  I will fix!

When we set the bounds for the cut scene animation, the user can no longer control the player so I think it's ok.

joshlory: Note that `this.game.world.setBounds` only constrains the camera.  Sprites don't take the bounds into account when moving.

joshlory:
@Hamms PTAL!

PR #94, opened 2017-09-11
Piston pr

WilliamHarmonMS:
Need to add unit test, and will be doing that right after my 1pm meeting!


src/js/game/LevelMVC/LevelView.js
```diff
@@ -1806,6 +1822,9 @@ module.exports = class LevelView {
         break;
 
       default:
+        if (this.blocks[blockType] === undefined) {
+          console.log("what the fuck?");
```
joshlory: Extra?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -325,9 +325,144 @@ module.exports = class LevelPlane extends Array {
       posToRefresh.push(this.redstoneListON[i]);
     }
 
+    //once we're done updating redstoneWire states, check to see if doors should open/close
```
joshlory: This method is getting pretty complex — can we start separating out concerns into different functions?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -325,9 +325,144 @@ module.exports = class LevelPlane extends Array {
...
+      }
+
+      if (this[i].blockType.startsWith("piston")) {
+        this[i].isPowered = this.powerCheck(this.indexToCoordinates(i));
+        if (this[i].isPowered) {
+          this.activatePiston(this.indexToCoordinates(i));
+          this.levelModel.controller.updateFowPlane();
+          this.levelModel.controller.updateShadingPlane();
+        } else if (!this[i].isPowered) {
+          this.deactivatePiston(this.indexToCoordinates(i));
```
joshlory: We're calling `this.indexToCoordinates(i)` repeatedly in this added code.  Can we convert once and work with `position` instead of index?

Better yet, let's iterate through the positions so we're not directly dealing with indices until we need to actually look up items in the array.  See my feedback here: https://github.com/code-dot-org/craft/pull/67/files#r137057705.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -325,9 +325,144 @@ module.exports = class LevelPlane extends Array {
...
+          this.setBlockAt(neighborPosition[2], emptyBlock, false, true);
+        }
+        break;
+      }
+      case "pistonLeft": {
+        if (this[this.coordinatesToIndex(neighborPosition[3])].blockType === "pistonArm") {
+          this.setBlockAt(neighborPosition[3], emptyBlock, false, true);
+        }
+        break;
+      }
```
joshlory: Can we generalize these similar 4 branches of code?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -373,4 +508,18 @@ module.exports = class LevelPlane extends Array {
     }
   }
 
+  /**
+  * Checking power state for objects that are powered by redstone.
+  */
+  powerCheck(position) {
```
joshlory: Same feedback as https://github.com/code-dot-org/craft/pull/78#discussion_r137634552 here.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -87,11 +87,11 @@ module.exports = class LevelPlane extends Array {
   * Changes the block at a desired position to the desired block.
   * Important note: This is the cornerstone of block placing/destroying.
   */
-  setBlockAt(position, block) {
+  setBlockAt(position, block, force = false, skipRedstone = false) {
```
joshlory: Not used?  The `force = false` param looks like it might be the result of a bad merge.

joshlory: Please add `@param {boolean} skipRedstone Description of what the param does...` to the JSDoc comment.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -87,11 +87,11 @@ module.exports = class LevelPlane extends Array {
...
     this[this.coordinatesToIndex(position)] = block;
 
     let redstoneToRefresh = [];
-    if (block.isRedstone || block.blockType === '' || block.isConnectedToRedstone) {
+    if (!skipRedstone && (block.isRedstone || block.blockType === '' || block.isConnectedToRedstone)) {
```
joshlory: Why do these updates skip redstone propagation?  Can pistons move blocks (or entities) onto a piston that triggers a redstone update?

test/unit/LevelPlaneTest.js
```diff
@@ -433,3 +433,69 @@ test('torch charge: place block', t => {
...
+  expected.levelModel = null;
+  expected.redstoneList = [];
+  expected.redstoneListON = [];
+
+  t.deepEqual(plane.map(block => block.blockType), expected);
+
+  t.end();
+});
+
+test('piston deactivate: block block', t => {
```
joshlory: Tiny nit: "destroy block" ? 😄

test/unit/LevelPlaneTest.js
```diff
@@ -433,3 +433,69 @@ test('torch charge: place block', t => {
...
+    '','','pistonUp','redstoneWireUpLeftOn','','redstoneWireUpRightOn','pistonDown','',
+    '','','','','','','pistonArm','',
+    '','','','','','','grass','',
+  ];
+
+  expected.width = undefined;
+  expected.height = undefined;
+  expected.levelModel = null;
+  expected.redstoneList = [];
+  expected.redstoneListON = [];
```
joshlory: Why is the `redstoneListON` empty here?

test/unit/LevelPlaneTest.js
```diff
@@ -433,3 +433,69 @@ test('torch charge: place block', t => {
...
+
+  t.deepEqual(plane.map(block => block.blockType), expected);
+
+  t.end();
+});
```
joshlory: Tests look good! 👍

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -325,9 +325,148 @@ module.exports = class LevelPlane extends Array {
...
+    }
+  }
+
+  pushBlocks(blocksPositions, offset) {
+    let armBlock = new LevelBlock("pistonArm");
+    // Need to check if the last index is up against the edge of the map, and break out if that's the case.
+    if (!this.inBounds([blocksPositions[blocksPositions.length - 1][0] + offset[0], blocksPositions[blocksPositions.length - 1][1] + offset[1]])) {
+      // Apparently me might NOT want this check? Or might? Design undecided.
+      //break;
+    }
```
joshlory: Not used?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -325,9 +325,148 @@ module.exports = class LevelPlane extends Array {
...
+    for (let i = blocksPositions.length - 1; i >= 0; --i) {
+      let destination = [blocksPositions[i][0] + offset[0], blocksPositions[i][1] + offset[1]];
+      this.setBlockAt(destination, this.getBlockAt(blocksPositions[i]));
+      if (i === 0) {
+        this.setBlockAt(blocksPositions[i], armBlock);
+      }
+    }
+  }
+
+  getBlocksToPush(position, offset) {
```
joshlory: Maybe worth matching the signature of `getBlockAt` where `offsetX` and `offsetY` are passed in separately?

```js
getBlockAt(position, offsetX = 0, offsetY = 0) { ... }
```

It's not clear to me that either one is better than the other, but my vote would be for them to be consistent 😄.

test/unit/LevelPlaneTest.js
WilliamHarmonMS: Um... what? Are you telling me you haven't been doing block block testing? Jiminy Crickets...

test/unit/LevelPlaneTest.js
WilliamHarmonMS: Because I'm not evaluating the redstone wires themselves. Just the piston states. That list would have all the wire that's powered, which we could evaluate, but as long as the Pistons are getting their state updated correctly, that's all I was concerned with.

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: Ah, yeah. I didn't know exactly what to do with that. Let me go see if Design input can still be found.

test/unit/LevelPlaneTest.js
WilliamHarmonMS: 👍

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: That's a good point. Will change 👍

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: Confirmed taking this out.

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: I was overflowing the stack because evaluate redstone -> evaluate piston -> pushes blocks with setBlockAt() which -> re-evaluates redstone etc etc.

It's basically a case where I was changing a block that was connected to redstone but I didn't want to re-evaluate all the redstone. I can look into making this cleaner if need be.

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: oh, yeah the force = false I don't even know if that was me? I definitely don't need it.

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: Haaaaaaaa clearly I was unamused by undefined block types 👎

src/js/game/LevelMVC/LevelPlane.js
Hamms: I'm definitely in favor of making this cleaner; I think we're opening ourselves up to some inconsistent/unpredictable redstone update behavior with this.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -325,10 +326,168 @@ module.exports = class LevelPlane extends Array {
...
+      }
+      case "pistonLeft": {
+        if (neighbors.west.block.blockType !== "" && neighbors.west.block.blockType !== "pistonArm") {
+          let offset = [-1,0];
+          let blocksPositions = this.getBlocksToPush(neighborPosition[3], offset[0], offset[1]);
+          this.pushBlocks(blocksPositions, offset[0], offset[1]);
+        } else if (neighbors.west.block.blockType === "") {
+          let armBlock = new LevelBlock("pistonArm");
+          this.setBlockAt(neighborPosition[3], armBlock);
+        }
```
Hamms: This is a lot of repeated code; I think it would be a lot nicer to just assign `neighbor`, `position`, and `offset` in the case block and then just do

        if (neighbor.block.blockType !== "" && neighbor.block.blockType !== "pistonArm" ) {
          let blocksPositions = this.getBlocksToPush(neighborPosition[position], offset[0], offset[1]);
          this.pushBlocks(blocksPositions, offset[0], offset[1]);
        } else if (neighbor.block.blockType === "") {
          let armBlock = new LevelBlock("pistonArm");
          this.setBlockAt(neighborPosition[position], armBlock);
        }

once afterwards

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1179,6 +1185,16 @@ module.exports = class LevelView {
   }
 
   refreshActionPlane(positions) {
+    // We need to add indices to refresh if there are other blocks in the action plane that might
+    // conflict with the drawing of refreshed blocks.
+    for (let i = 0; i < positions.length; ++i) {
+      if (positions[i][1] + 1 < this.controller.levelModel.actionPlane.height &&
+      this.controller.levelModel.actionPlane[this.controller.levelModel.actionPlane.coordinatesToIndex([positions[i][0], positions[i][1] + 1])].blockType !== "") {
```
Hamms: nit: could you break these out into clearly-named variables for clarity?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -325,10 +326,168 @@ module.exports = class LevelPlane extends Array {
...
+      case "pistonDown": {
+        this.retractArm(neighborPosition[1]);
+        break;
+      }
+      case "pistonRight": {
+        this.retractArm(neighborPosition[2]);
+        break;
+      }
+      case "pistonLeft": {
+        this.retractArm(neighborPosition[3]);
```
Hamms: Likewise here; it would be much nicer to have a standard way of translating the various piston blockType into the cardinal directions, then simply reference that here once as something like `this.retractArm(neighborPosition[direction]);`

WilliamHarmonMS: At the moment, the getOrthogonalPositions just returns an array of 4 2-element-arrays (positions). Thus the numeric indexing. I mean, the """direction""" that controller uses in moveDirection() is the same way (just 0-4 as an int). Which is awful. 

we COULD refactor getOrthogonalPositions to be like getOrthogonalBlocks and return an object that has 4 elements that are other objects called north/south/east/west so that we could reference neighborPosition.north? I don't know how extensive of a refactor that would be though, given that getOrthogonalPositions is used in a few areas?

Let me know what trigger you want to pull on this one.

WilliamHarmonMS: Or I could just #define north 0;

😂 😂 😂 😂

WilliamHarmonMS: Curses! @joshlory beat me to that idea!

const North = 1;
const South = 2;
const East = 4;
const West = 8;

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: 👍

src/js/game/LevelMVC/LevelPlane.js
joshlory: This PR has some convenience methods that may help?  https://github.com/code-dot-org/craft/pull/96/files#r138232605

I made a big mistake by adding North/South/East/West.  I'm planning to unify with Up/Down/Right/Left!

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: What do you mean here? positions is an array of [x,y] elements, and that loop is finding if any element that exists in position has another element below it (because refreshing something that has an overlap from a southern index causes a draw conflict for like 1 frame) and then adds that southern element to the list-to-refresh (if you refresh both, the draw conflict doesn't manifest).

I can take an extra step to clarify it a bit, but do let me know if you have something specific in mind 👌

Hamms: Something along the lines of

```javascript
const positionBelow = [positions[i][0], positions[i][1] + 1];
const isPositionBelowValid = positionBelow[1] < this.controller.levelModel.actionPlane.height;
if (!isPositionBelowValid) {
  continue;
}

const indexBelow = this.controller.levelModel.actionPlane.coordinatesToIndex(positionBelow)
const isBlockBelow = this.controller.levelModel.actionPlane[indexBelow].blockType !== ""
if (isBlockBelow) {
  positions.push(positionBelow);
}
```

would be grand

WilliamHarmonMS: Ahhh alright, I can break this up even more. I stopped a few short of that, just explicitly defining that we were looking at blocks.

Hamms: For a codebase like this one that gets looked at once a year by a team that might not be the same as the team that looked at it last year, there's no such thing as making your code _too_ explicit

WilliamHarmonMS: Roger that. I broke up all the pieces that looked rationally break-up-able? It reads pretty clear to me, but again I am more than happy to change anything else.

Hamms: looks great!

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -86,12 +86,13 @@ module.exports = class LevelPlane extends Array {
   /**
   * Changes the block at a desired position to the desired block.
   * Important note: This is the cornerstone of block placing/destroying.
+  * @param {boolean} skipRedstone: if for some reason we want to place a block and bypass the redstone evaluation.
```
Hamms: nit: outdated comment

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -325,10 +326,164 @@ module.exports = class LevelPlane extends Array {
...
+          this.levelModel.controller.updateFowPlane();
+          this.levelModel.controller.updateShadingPlane();
+        }
+      } else if (!this[index].isPowered) {
+        this.deactivatePiston(this.indexToCoordinates(index));
+        if (this.levelModel) {
+          this.levelModel.controller.updateFowPlane();
+          this.levelModel.controller.updateShadingPlane();
+        }
+      }
```
Hamms: This can be simplified to:

```javascript
const coordinates = this.indexToCoordinates(index)

if (this[index].isPowered) {
  this.activatePiston(coordinates);
} else {
  this.deactivatePiston(coordinates);
}

if (this.levelModel) {
  this.levelModel.controller.updateFowPlane();
  this.levelModel.controller.updateShadingPlane();
}
```

Hamms: Although that does raise the question of whether or not we want to do all this work every time or only when `isPowered` changes

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -325,10 +326,164 @@ module.exports = class LevelPlane extends Array {
...
+        break;
+      }
+    }
+    if (workingNeighbor.blockType !== "" && workingNeighbor.blockType !== "pistonArm") {
+      let blocksPositions = this.getBlocksToPush(pos, offset[0], offset[1]);
+      this.pushBlocks(blocksPositions, offset[0], offset[1]);
+    } else if (workingNeighbor.blockType === "") {
+      let armBlock = new LevelBlock("pistonArm");
+      this.setBlockAt(pos, armBlock);
+    }
```
Hamms: :+1: much cleaner!

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -105,7 +105,7 @@ module.exports = class LevelPlane {
     }
 
     let redstoneToRefresh = [];
-    if (block.isRedstone || block.blockType === '' || block.isConnectedToRedstone) {
+    if (block.isRedstone || block.blockType === '' || (block.isConnectedToRedstone && !block.blockType.startsWith("piston"))) {
```
Hamms: this conditional is becoming a bit unwieldy, and it's not clear just by looking at it why pistons don't count. I vote we move this whole conditional into a method on LevelBlock that's well-named and -commented.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -265,24 +265,70 @@ module.exports = class LevelPlane {
...
+  /**
+  * Evaluates what state Pistons on the map should be in.
+  */
+  getPistonState(index) {
+    if (this._data[index].blockType.startsWith("piston") && !this._data[index].blockType.startsWith("pistonArm")) {
+      this._data[index].isPowered = this.powerCheck(this.indexToCoordinates(index));
+      if (this._data[index].isPowered) {
+        this.activatePiston(this.indexToCoordinates(index));
+        if (this.levelModel) {
+        }
```
Hamms: nit: empty block

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -265,24 +265,70 @@ module.exports = class LevelPlane {
...
+        this[i].isPowered = this.powerCheck(this.indexToCoordinates(i));
+        if (this[i].isPowered && !this[i].isOpen) {
+          this[i].isOpen = true;
           if (this.levelModel) {
             this.levelModel.controller.levelView.animateDoor(i, true);
           }
-        } else if (!this._data[i].isPowered && this._data[i].isOpen) {
-          this._data[i].isOpen = false;
+        } else if (!this[i].isPowered && this[i].isOpen) {
+          this[i].isOpen = false;
```
Hamms: this should all be using `_data`

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -292,6 +338,147 @@ module.exports = class LevelPlane {
...
+
+  /**
+  * Deactivates a piston at a given position by determining what the arm orientation is.
+  */
+  deactivatePiston(position) {
+    let neighborPosition = this.getOrthogonalPositions(position);
+    let north = 0;
+    let south = 1;
+    let east = 2;
+    let west = 3;
```
Hamms: nit: should just use the existing FacingDirection constants for this

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1184,6 +1198,22 @@ module.exports = class LevelView {
   }
 
   refreshActionPlane(positions) {
+    // We need to add indices to refresh if there are other blocks in the action plane that might
+    // conflict with the drawing of refreshed blocks.
+    for (let i = 0; i < positions.length; ++i) {
+      const positionBelow = [positions[i][0], positions[i][1] + 1];
+      const indexIsValid = this.controller.levelModel.actionPlane.inBounds(positionBelow);
+      if (indexIsValid) {
+        const positionOfInterest = [positions[i][0], positions[i][1] + 1];
```
Hamms: nit: why not reuse `positionBelow`?

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: Ah, I see that that's been reworked? Initially the values at the top of LevelPlane were bit shifted values (1,2,4,8) not subsequent index values (0,1,2,3). I opted to just define there than do math to get them into index values. If this has been changed I'll gladly use the new values 😄

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: d'oh

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: 👍

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: Becaaause there was another step and then I re-wrote some of it, and didn't realize that the logic became duplicated 😦

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: Put a comment into the function to clean this up later.

PR #96, opened 2017-09-12
Minecart movement

joshlory:
Initial minecart movement for arbitrary track layouts.

minecart-movement

I don't handle entering or leaving a minecart yet, only movement when already on a set of tracks.  You can manually start the animation by standing on a minecart track and running:

```js
gameController.levelView.playTrack(gameController.player.position, gameController.player.facing);
```


src/js/game/LevelMVC/FacingDirection.js
```diff
@@ -1,6 +1,50 @@
...
+  turnDirection: function (from, to) {
+    switch (from) {
+      case FacingDirection.North: return to === FacingDirection.East ? 'right' : 'left';
+      case FacingDirection.South: return to === FacingDirection.West ? 'right' : 'left';
+      case FacingDirection.East: return to === FacingDirection.South ? 'right' : 'left';
+      case FacingDirection.West: return to === FacingDirection.North ? 'right' : 'left';
+    }
+  },
+
+  // TODO: unify with N/S/E/W above.
```
joshlory: I made a huge mistake by adding North/South/East/West instead of re-using Up/Down/Right/Left. 😦

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -146,6 +151,43 @@ module.exports = class LevelPlane extends Array {
...
+    if (block.connectionA === incomming && block.connectionB) {
+      const rotation = turnDirection(cardinal, block.connectionB);
+      const newFacing = turn(facing, rotation);
+      return [`turn_${rotation}`, position, newFacing, speed];
+    }
+    if (block.connectionB === incomming && block.connectionA) {
+      const rotation = turnDirection(cardinal, block.connectionA);
+      const newFacing = turn(facing, rotation);
+      return [`turn_${rotation}`, position, newFacing, speed];
+    }
```
joshlory: I think I can extract the common code from these two cases.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -146,6 +151,43 @@ module.exports = class LevelPlane extends Array {
...
+    const [x, y] = position;
+    switch (cardinal) {
+      case North: return [x, y - 1];
+      case South: return [x, y + 1];
+      case East: return [x + 1, y];
+      case West: return [x - 1, y];
+    }
+  }
+
+  getMinecartTrack(position, facing) {
```
Hamms: I'd love to see some documentation for the return value parameters

src/js/game/LevelMVC/FacingDirection.js
Hamms: Having both concepts is definitely confusing. How much work will it be to unify them?

src/js/game/LevelMVC/LevelView.js
```diff
@@ -536,48 +537,47 @@ module.exports = class LevelView {
...
-    var animation;
-    this.track = minecartTrack;
-    this.i = 0;
-
+  playMinecartAnimation(isOnBlock, completionHandler, unpoweredRails) {
     //start at 3,2
     this.setPlayerPosition(3, 2, isOnBlock);
-    position = [3, 2];
+    const position = [3, 2];
+    this.player.facing = 2;
```
Hamms: I'm surprised to still see this here; shouldn't it be generalized away as a part of this PR?

src/js/game/LevelMVC/FacingDirection.js
joshlory: It shouldn't be too bad — I definitely want to do it before merging this PR.

src/js/game/LevelMVC/LevelView.js
joshlory: Unfortunately this animation is specific to the Adventurer 13 minecart level.  The player enters the house facing north, but in the cut scene animation we move them to location [3, 2] and make them face south.  Then we start calling `playTrack` to make them drive off screen.

joshlory:
@Hamms This change includes PR https://github.com/code-dot-org/craft/pull/96.  After merging 96 to `master`, merge `master` into this branch to reduce the diff of this PR.

Thanks!

PR #100, opened 2017-09-12
Reset should move the camera to start position (when not following)

joshlory:
Fixes running Adventurer 13 a second time, after the cut scene animation has finished.


joshlory:
The issue before this fix.  Resetting the level doesn't move the camera back to the start position:

image

PR #106, opened 2017-09-13
Facing direction cleanup

joshlory:
Address https://github.com/code-dot-org/craft/pull/96#discussion_r138232605.


joshlory:
@Hamms If you merge this change, you'll also need to update `designer/craft.js` when integrating upstream: https://github.com/code-dot-org/code-dot-org/search?q=FacingDirection

PR #107, opened 2017-09-13
add support for miniblocks in level specification

Hamms:
Levelbuilders can now add miniblocks that appear at level initialization, and which will be collected by the player just like other collectables:

diamond

Two things to note:

First, the default range for collectables is pretty large; the feature this functionality is desired for involves adding diamonds to levels to give the player an extra challenge, and it would probably be nice for that to have a more restrictive area. (but that's an issue for a separate PR)

Second, to enable this on levelbuilder, we simply need to add the appropriate strings to the set of block options for the level:

```
diff --git a/dashboard/app/models/levels/craft.rb b/dashboard/app/models/levels/craft.rb
index 6a551e026f..b8300b3cc5 100644
--- a/dashboard/app/models/levels/craft.rb
+++ b/dashboard/app/models/levels/craft.rb
@@ -221,6 +221,7 @@ class Craft < Blockly
     },
     action_plane: ALL_BLOCKS.merge(
       {
+        diamondMiniblock: true,
         creeper: true,
         sheep: true,
         cropWheat: true,
```

### DIFF
--- a/src/js/game/LevelMVC/LevelBlock.js
+++ b/src/js/game/LevelMVC/LevelBlock.js
@@ -146,8 +146,11 @@ module.exports = class LevelBlock {
     }
   }
 
+  /**
+   * @return {boolean}
+   */
   getIsMiniblock() {
-    return this.blockType.endsWith("Miniblock")
+    return LevelBlock.isMiniblock(this.blockType);
   }
 
   getIsTree() {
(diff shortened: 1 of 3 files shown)


src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -152,4 +168,94 @@ module.exports = class LevelBlock {
...
+   * block will produce a "diamond" frame.
+   *
+   * For regular block types, this should be the miniblock produced when
+   * destroying the block type, so a "oreDiamond" block will produce a "diamond"
+   * frame
+   *
+   * @param {String} blockType
+   * @return {String} frame identifier
+   */
+  static getMiniblockFrame(blockType) {
```
joshlory: 👏

PR #108, opened 2017-09-14
just getting started: introducing Agent.js

WilliamHarmonMS:


src/js/game/Entities/Agent.js
```diff
@@ -0,0 +1,156 @@
+const BaseEntity = require("./BaseEntity.js");
+const CallbackCommand = require("../CommandQueue/CallbackCommand.js");
+
+module.exports = class Player extends BaseEntity {
```
joshlory: There's a lot of duplicated functionality here from Player.js.  Can you have `class Agent extends Player` and only override the methods that behave differently?

src/js/game/Entities/Agent.js
```diff
@@ -0,0 +1,156 @@
...
+  constructor(controller, type, x, y, name, isOnBlock, facing) {
+    super(controller, type, 'PlayerAgent', x, y, facing);
+    this.offset = (name === 'Agent') ? [-5, -15] : [-18, -32];
```
joshlory: Agent offset is always `[-5, -15]`.

src/js/game/Entities/Agent.js
```diff
@@ -0,0 +1,156 @@
...
+      levelView = this.controller.levelView;
+    let wasOnBlock = player.isOnBlock;
+    let prevPosition = this.position;
+    // update position
+    levelModel.moveForward();
+    // TODO: check for Lava, Creeper, water => play approp animation & call commandQueueItem.failed()
+
+    jumpOff = wasOnBlock && wasOnBlock !== player.isOnBlock;
+    if (player.isOnBlock || jumpOff) {
+      groundType = levelModel.actionPlane._data[levelModel.yToIndex(player.position[1]) + player.position[0]].blockType;
```
joshlory: Please use `levelModel.actionPlane.getBlockAt(position)` (one more below).

src/js/game/Entities/Player.js
```diff
@@ -4,7 +4,7 @@ const CallbackCommand = require("../CommandQueue/CallbackCommand.js");
 module.exports = class Player extends BaseEntity {
   constructor(controller, type, x, y, name, isOnBlock, facing) {
     super(controller, type, 'Player', x, y, facing);
-    this.offset = (name === 'Agent') ? [-5, -15] : [-18, -32];
+    this.offset = (name === 'Player') ? [-5, -15] : [-18, -32];
```
joshlory: This will set the Player offset to `[-5, -15]` which isn't correct.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1462,6 +1497,69 @@ module.exports = class LevelView {
...
+    this.agent.sprite.animations.add('mineCart_turnright_down', Phaser.Animation.generateFrameNames("Minecart_", 12, 12, "", 2), frameRate, false);
+
+    this.agent.sprite.animations.add('mineCart_right', Phaser.Animation.generateFrameNames("Minecart_", 7, 7, "", 2), frameRate, false);
+    this.agent.sprite.animations.add('mineCart_left', Phaser.Animation.generateFrameNames("Minecart_", 11, 11, "", 2), frameRate, false);
+
+    this.agent.sprite.animations.add('mineCart_up', Phaser.Animation.generateFrameNames("Minecart_", 9, 9, "", 2), frameRate, false);
+    this.agent.sprite.animations.add('mineCart_turnleft_up', Phaser.Animation.generateFrameNames("Minecart_", 10, 10, "", 2), frameRate, false);
+    this.agent.sprite.animations.add('mineCart_turnright_up', Phaser.Animation.generateFrameNames("Minecart_", 8, 8, "", 2), frameRate, false);
+  }
+
```
joshlory: Why is this duplicating code from `preparePlayerSprite`?  https://github.com/code-dot-org/craft/blob/8fa65f72d375f4d33aa376a99bcff3a84fc9a920/src/js/game/LevelMVC/LevelView.js#L1434-L1498

src/js/game/Entities/Agent.js
WilliamHarmonMS: Yeah, that's the end goal. At the moment, I just copied it over with a different tag, trying to get both entities to behave at the same time. A clean up would absolutely entail extending Player.js and then only overloading necessary functions.

I would expect that to be done before the PR is accepted.

src/js/game/Entities/Agent.js
WilliamHarmonMS: Any idea what we want the offset to be?

src/js/game/Entities/Agent.js
WilliamHarmonMS: Yup 👍 
again, that's just copy pasted from player.js
Once functionality is working I'm going to clean these files up.

src/js/game/Entities/Player.js
WilliamHarmonMS: is [-18, -32] the desired offset? When is the name anything other than Player? I didn't see the need for this offset shuffle, but perhaps I just haven't seen where it comes into play?

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: Setting up the the animations for this.agent?

this.agent.sprite = this.actionPlane.create(0, 0, `player${playerName}`, 'Player_121');
^With the string playerAgent it should be referencing the agent JSON ~> Agent png and then adding all the sprites (which have the same names as the Alex/Steve sprites I believe) to the agent's sprite.animations.

That was my objective anyway.

WilliamHarmonMS: Also @joshlory aren't you on vacation? Why are you responding to work stuff? 😂

src/js/game/Entities/Agent.js
Hamms: Seconded that this should definitely be a subclass

src/js/game/LevelMVC/LevelView.js
Hamms: Dammit Josh, go relax!

joshlory: :smile: :palm_tree:

joshlory: Right now preparePlayerSprie handles both cases (Agent, Steve, Alex, SteveEvents, AlexEvents). We shouldn't need to duplicate the code in order to keep the Agent working.

src/js/game/Entities/Player.js
joshlory: I could be mistaken -- I think [-18, -32] is correct for the player. That was the value before I added [-5, -15] for the Agent.

src/js/game/Entities/Agent.js
joshlory: Sounds good!

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: Yeah, the logic in preparePlayerSprite logic is 100% covered, but it directly references this.player (which is just levelModel.player) when assigning sprites. If we're going to have a whole second entity we'll need to be able to assign sprites to that entity as well. I'm making the function generalized this morning to take an optional entity argument (default to this.player) that should eliminate the need to make a duplicate function 👍

src/js/game/Entities/Agent.js
WilliamHarmonMS: Also, just noticed.... this was copied over from Player.js but is this if/else really:
if(){
 groundType = levelModel.actionPlane.getBlockAt(player.position).blockType;
} else {
 groundType = levelModel.actionPlane.getBlockAt(player.position).blockType;
}

#ForWhatPurpose

Hamms: Oh, the else block is supposed to be referencing the _ground_ plane, not the _action_ plane

joshlory:
Test coverage?

Why add `usingAgent: false`, `agentStartPosition: [3, 7]`, and `agentStartDirection: 0` to all the Designer levels instead of making this the default?

PR #109, opened 2017-09-14
Add levelPlace._data access methods

Hamms:
In response to https://github.com/code-dot-org/craft/pull/99

Added a `getAllPositions` method for operations which want to examine all blocks on a plane, and updated everything to use `getBlockAt` positional access rather than index access.

Note that this required updating in a couple of places the way we're approaching an operation, and in a couple of other places it requires awkwardly converting indexes to positions. I think for the former considerations all the changes resulted in improved code, and in the latter consideration it serves to highlight the fact that we would probably prefer to be using positions more universally.

### DIFF
--- a/src/js/game/GameController.js
+++ b/src/js/game/GameController.js
@@ -1281,7 +1281,7 @@ class GameController {
     if (!this.levelModel.inBounds(position[0], position[1])) {
       return;
     }
-    let block = this.levelModel.actionPlane.getBlock(this.levelModel.yToIndex(position[1]) + position[0]);
+    let block = this.levelModel.actionPlane.getBlockAt(position);
 
     if (block !== null && block !== undefined) {
       let destroyPosition = position;
@@ -1349,7 +1349,7 @@ class GameController {
         this.levelModel.destroyBlock(blockIndex);
       }
 
-      if (blockType !== "cropWheat" || this.levelModel.groundPlane.getBlockAt((this.levelModel.player.position)).blockType === "farmlandWet") {
+      if (blockType !== "cropWheat" || this.levelModel.groundPlane.getBlockAt(this.levelModel.player.position).blockType === "farmlandWet") {
         this.levelModel.player.updateHidingBlock(this.levelModel.player.position);
         this.levelView.playPlaceBlockAnimation(this.levelModel.player.position, this.levelModel.player.facing, blockType, blockTypeAtPosition, () => {
           if (this.checkMinecartLevelEndAnimation() && blockType === "rail") {
--- a/src/js/game/GameController.js
+++ b/src/js/game/GameController.js
@@ -1160,8 +1160,8 @@ class GameController {
     let player = this.levelModel.player;
     let frontPosition = this.levelModel.getMoveForwardPosition(player);
     let frontEntity = this.levelEntity.getEntityAt(frontPosition);
-    const frontIndex = this.levelModel.yToIndex(frontPosition[1]) + frontPosition[0];
-    let frontBlock = this.levelModel.actionPlane.getBlock(frontIndex);
+    let frontBlock = this.levelModel.actionPlane.getBlockAt(frontPosition);
+
     const isFrontBlockDoor = frontBlock === undefined ? false : frontBlock.blockType === "door";
     if (frontEntity !== null) {
       // push use command to execute general use behavior of the entity before executing the event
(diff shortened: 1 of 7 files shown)


src/js/game/LevelMVC/LevelModel.js
```diff
@@ -1175,30 +1175,30 @@ module.exports = class LevelModel {
...
-          if (x < this.planeWidth - 1 && !this.actionPlane._data[this.yToIndex(y + 1) + x + 1].getIsEmptyOrEntity() &&
-            this.actionPlane._data[this.yToIndex(y) + x + 1].getIsEmptyOrEntity()) {
+          if (x < this.planeWidth - 1 && !this.actionPlane.getBlock(this.yToIndex(y + 1) + x + 1).getIsEmptyOrEntity() &&
+            this.actionPlane.getBlock(this.yToIndex(y) + x + 1).getIsEmptyOrEntity()) {
             // needs a bottom left side AO shadow
             this.shadingPlane.push({ x: x, y: y, type: 'AOeffect_TopLeft' });
           }
 
-          if (!hasRight && x > 0 && !this.actionPlane._data[this.yToIndex(y + 1) + x - 1].getIsEmptyOrEntity()) {
+          if (!hasRight && x > 0 && !this.actionPlane.getBlock(this.yToIndex(y + 1) + x - 1).getIsEmptyOrEntity()) {
```
Hamms: note that these could probably be updated to use `getBlockAt` with the offsets already provided by that method, but that's slightly out of scope for this PR

joshlory: I think we probably want to abstract away the underlying array that's storing the data. Consumers shouldn't know/care that the blocks are stored as a flat array vs an array of arrays vs some other data structure.

Is there anywhere we're accessing blocks by index that can't be written more cleanly with position?

Hamms: There are mostly a couple of places where we don't care and just want to be able to iterate through all the blocks in some order; there are also plenty of other places that are still using index internally for other purposes and which would require converting from index to position if we wanted to only support `getBlockAt` and never just `getBlock`.

I agree that I'd like to move more in the direction of just using position everywhere, but this feels like a good place for us to be at for now

Hamms: Alternatively, we could just add a method to get all the blocks for those iteration purposes and then update all the other places that are using index internally to convert to position, thus doing a better job of internally communicating that we really actually want to just always be using position.

The more I think about it the more I think I prefer that.

src/js/game/GameController.js
```diff
@@ -1280,14 +1280,15 @@ class GameController {
...
   placeBlock(commandQueueItem, blockType) {
-    let blockIndex = (this.levelModel.yToIndex(this.levelModel.player.position[1]) + this.levelModel.player.position[0]);
-    let blockTypeAtPosition = this.levelModel.actionPlane._data[blockIndex].blockType;
+    const position = this.levelModel.player.position;
+    let blockTypeAtPosition = this.levelModel.actionPlane.getBlockAt(position).blockType;
+
     if (this.levelModel.canPlaceBlock()) {
       if (blockTypeAtPosition !== "") {
-        this.levelModel.destroyBlock(blockIndex);
+        this.levelModel.destroyBlock(position);
```
Hamms: note that destroyBlock expects to be passed a position argument, so I'm pretty sure this wasn't working correctly before

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -35,20 +35,17 @@ module.exports = class LevelModel {
...
-        let position = this.actionPlane.indexToCoordinates(i);
+    this.actionPlace.getAllPositions().forEach((position) => {
+      if (this.actionPlane.getBlockAt(position).blockType === "railsRedstoneTorch") {
         this.actionPlane.redstonePropagation(position);
       }
-    }
-    for (let i = 0; i < this.actionPlane._data.length; ++i) {
-      if (this.actionPlane._data[i].blockType.substring(0,12) === "redstoneWire") {
-        let y = Math.floor(i / this.planeHeight);
-        let x = i - (y * this.planeHeight);
```
Hamms: note that this was previously using `planeHeight` when I'm pretty sure we actually want x and y to be determined by planeWidth

src/js/game/LevelMVC/LevelModel.js
Hamms: Okay, yep, I like that _much_ better. `getBlock` is dead, long live `getBlockAt`

PR #111, opened 2017-09-18
Agent can now walk over water and lava

WilliamHarmonMS:


src/js/game/LevelMVC/LevelModel.js
```diff
@@ -666,7 +667,7 @@ module.exports = class LevelModel {
         // there is no entity
         && (frontEntity === undefined)
         // no lava or water
-        && (this.groundPlane.getBlockAt(position).blockType !== "water" && this.groundPlane.getBlockAt(position).blockType !== "lava");
+        && (entity === this.agent || (this.groundPlane.getBlockAt(position).blockType !== "water" && this.groundPlane.getBlockAt(position).blockType !== "lava"));
```
Hamms: I'd much prefer to see something like `entity.canWalkOnBlock(this.groundPlane.getBlockAt(position))`, and have `player` and `agent` simply define the method themselves

WilliamHarmonMS: Roger that 👍 

I can do that first thing tomorrow.

src/js/game/Entities/Agent.js
```diff
@@ -153,4 +153,14 @@ module.exports = class Agent extends BaseEntity {
...
   }
+
+  hasPermissionToWalk(actionBlock, frontEntity, groundBlock = null) {
+        return (actionBlock.isWalkable || ((frontEntity !== undefined && frontEntity.isOnBlock)
+        // action plane is empty
+        && !actionBlock.isEmpty))
+        // there is no entity
+        && (frontEntity === undefined)
+        //linter won't let groundBlock go unused, even though the agent doesn't need it. So...
+        && (groundBlock === null || groundBlock !== null);
```
Hamms: You can actually just remove the argument from this method definition, or rename it to `_`

WilliamHarmonMS: LevelModel calls this function like:
result[0] = entity.hasPermissionToWalk(actionBlock, frontEntity, groundBlock);

I didn't want to make a special case just for if(entity==agent) just so I could pass only two arguments?

Hamms: oh, you don't have to.

```javascript
function foo (bar, baz) {
  console.log(bar, baz);
}

foo(1)

>> 1 undefined
```

is totally valid

Hamms: More relevant to your specific case, so is

```javascript
function foo (bar) {
  console.log(bar)
}

foo(1, 2, 3)

>> 1
```

joshlory:
Test coverage?

PR #113, opened 2017-09-19
Agent can destroy blocks and properly animates

WilliamHarmonMS:


src/js/game/LevelMVC/LevelView.js
```diff
@@ -879,14 +879,14 @@ module.exports = class LevelView {
     this.playScaledSpeed(destroyOverlay.animations, "destroy");
   }
 
-  playDestroyBlockAnimation(playerPosition, facing, destroyPosition, blockType, completionHandler) {
+  playDestroyBlockAnimation(entity, playerPosition, facing, destroyPosition, blockType, completionHandler) {
```
Hamms: We've started to develop a pattern of having `entity` be the last argument to a method, and having it default to player if appropriate. I'd like to maintain that pattern wherever we can

PR #119, opened 2017-09-20
piping hot biome assets

WilliamHarmonMS:


src/js/game/GameController.js
```diff
@@ -1307,6 +1307,8 @@ class GameController {
             break;
           case "logSpruce":
           case "treeSpruce":
+          case "logSpruceSnowyDecay":
+          case "treeSpruceSnowyDecay":
```
Hamms: this should also not be "Decay"

src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -137,6 +137,10 @@ module.exports = class LevelBlock {
       this.isRedstoneBattery = blockType === 'pressurePlateUp' ? false : true;
     }
 
+    if (blockType === "glowstone") {
+      this.isEntity = true;
```
Hamms: why is glowstone an entity?

src/js/game/LevelMVC/LevelView.js
```diff
@@ -104,6 +106,7 @@ module.exports = class LevelView {
       "leavesJungle": ["leavesJungle", "Leaves_Jungle0.png", -100, 0],
       "leavesOak": ["leavesOak", "Leaves_Oak0.png", -100, 0],
       "leavesSpruce": ["leavesSpruce", "Leaves_Spruce0.png", -100, 0],
+      "leavesSpruceSnowy": ["leavesSpruceSnowy", "Leaves_SpruceSnowy1.png", -100, 36],
```
Hamms: why not "Leaves_SpruceSnowy0.png"?

src/js/game/LevelMVC/LevelView.js
```diff
@@ -191,6 +194,18 @@ module.exports = class LevelView {
...
+      "dead_bush": ["blocks", "dead_bush", -13, 0],
+      "glowstone": ["blocks", "glowstone", -13, 0],
+      "grass_path": ["blocks", "grass_path", -13, 0],
+      "ice": ["blocks", "ice", -13, 0],
+      "netherrack": ["blocks", "netherrack", -13, 0],
+      "nether_brick": ["blocks", "nether_brick", -13, 0],
+      "quartz_ore": ["blocks", "quartz_ore", -13, 0],
+      "snow": ["blocks", "snow", -13, 0],
+      "snowy_grass": ["blocks", "snowy_grass", -13, 0],
+      "top_snow": ["blocks", "top_snow", -13, 0],
```
Hamms: I'm pretty sure at least some of these shouldn't be actual blocks, much like the "tallGrass" asset

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1691,6 +1706,9 @@ module.exports = class LevelView {
       case "treeSpruce": //0,8
         buildTree(this, [0, 8]);
         break;
+      case "treeSpruceSnowy": //1,9
+        buildTree(this, [1, 9]);
```
Hamms: why starting from 1?

WilliamHarmonMS: That was just the way the the json was named. For some reason this one came to me 1-9, not 0-8. That should be changeable if we cared, so long as it lined up. 🤷‍♂️

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: Same reason as below, I just followed the pattern.
Leaves_Spruce0.png is the first frame referenced on the Spruce png, and Leaves_SpruceSnowy1.png is the first frame referenced on the Spruce Snowy png.

src/js/game/GameController.js
WilliamHarmonMS: Good catch 👍

src/js/game/LevelMVC/LevelBlock.js
WilliamHarmonMS: Because entity doesn't mean entity. Blocks tagged as entity just mean "do not draw shadows / AO". Josh and I have spoken about how that's a really dumb name for how it's handled / what it effects. 😞

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: I'm actually of the opinion that -everything- in blocks[] should be a block... It would appear that the top half of everything is listed as a block (done in year 1) and then whoever did year 2 decided to start labeling things as "not blocks". 

Tall grass for instance, is basically the same as flowerRose (in that it's only meant to be used in the decorationPlane). flowerRose, as with all the flowers, are also given the key "blocks". 

The one sort of rational exception, in my opinion are entries with added behavior. Like tree leaves and doors, as they animate.

None of the additional entries in this PR, with the exception of "leavesSpruceSnowy" animate in any way, and are thus categorized as blocks like all the other non-animated entries.

Hamms: Huh, right you are. I had a misunderstanding on what it means to be not-a-block.

Thanks for the solid explanation!

src/js/game/LevelMVC/LevelBlock.js
joshlory: Can we update the name to reflect the new behavior?  Thanks!

PR #121, opened 2017-09-20
added a moveBackward command [COMPLETE]

WilliamHarmonMS:
Confirmed with design that a moveBackward() command is desired for optimal level solutions according to current curriculum. 

This implements the command, though the behavior is still identical to moveForward(). Here I'm just hooking up all the CommandQueue elements.


src/js/game/LevelMVC/LevelModel.js
```diff
@@ -599,6 +624,14 @@ module.exports = class LevelModel {
     return this.isPositionEmpty([x, y], entity);
   }
 
+  canMoveBackward(entity = this.player) {
+    const [x, y] = this.getMoveBackwardPosition(entity);
+    if (!this.controller.followingPlayer() && (x > 9 || y > 9)) {
```
Hamms: this condition is a custom check for a specific level, and doesn't need to be replicated in the `moveBackward` logic

src/js/game/GameController.js
```diff
@@ -784,6 +784,32 @@ class GameController {
...
 
+  moveBackward(commandQueueItem) {
+    let target = commandQueueItem.target;
+    if (!this.isType(target)) {
+      // apply to all entities
+      if (target === undefined) {
+        let entities = this.levelEntity.entityMap;
+        for (var value of entities) {
+          let entity = value[1];
+          let callbackCommand = new CallbackCommand(this, () => { }, () => { this.moveForward(callbackCommand); }, entity.identifier);
```
Hamms: this should be `this.moveBackward`

src/js/game/GameController.js
```diff
@@ -784,6 +784,32 @@ class GameController {
...
+        }
+        commandQueueItem.succeeded();
+      } else {
+        let entity = this.getEntity(target);
+        entity.moveBackward(commandQueueItem);
+      }
+    } else {
+      let entities = this.getEntities(target);
+      for (let i = 0; i < entities.length; i++) {
+        let callbackCommand = new CallbackCommand(this, () => { }, () => { this.moveForward(callbackCommand); }, entities[i].identifier);
```
Hamms: another `moveForward`

src/js/game/Entities/Agent.js
```diff
@@ -94,6 +94,49 @@ module.exports = class Agent extends BaseEntity {
...
+    // TODO: check for Lava, Creeper, water => play approp animation & call commandQueueItem.failed()
+
+    jumpOff = wasOnBlock && wasOnBlock !== player.isOnBlock;
+    if (player.isOnBlock || jumpOff) {
+      groundType = levelModel.actionPlane.getBlockAt(player.position).blockType;
+    } else {
+      groundType = levelModel.actionPlane.getBlockAt(player.position).blockType;
+    }
+
+    levelView.playMoveForwardAnimation(player, prevPosition, player.facing, jumpOff, player.isOnBlock, groundType, () => {
```
Hamms: since this method is being repurposed to be the animation for _all_ movement, it should also be renamed as such

src/js/game/LevelMVC/LevelModel.js
WilliamHarmonMS: Fair enough! I was just trying to replicate the whole line of logic to make sure it behaved the same. I'll remove this bit if it's not needed.

src/js/game/GameController.js
WilliamHarmonMS: Whoops. Copy-paste error. Good catch 👍

src/js/game/GameController.js
WilliamHarmonMS: Gah! Again. Copy-paste.

src/js/game/Entities/Agent.js
WilliamHarmonMS: That's an interesting point. Let me talk to Joel. The Agent has the same move animation regardless of the direction, but I think that Alex/Steve actually DO have slightly different animations if they're walking backwards. For future proofing, if we ever think Alex/Steve would use different animations, it might be worth it to split these up ahead of time?

Hamms: oh, even better! Yeah, let's go ahead and split these up

PR #125, opened 2017-09-20
Agent placeblock

WilliamHarmonMS:
Agent can place blocks beneath itself only if it is over water/lava

NOTE: [WIP] there are still animation problems.


src/js/game/LevelMVC/LevelModel.js
```diff
@@ -781,11 +781,19 @@ module.exports = class LevelModel {
         break;
     }
 
-    if (shouldPlace === true) {
-      var block = new LevelBlock(blockType);
+    if (entity === this.agent) {
+      let ground = this.groundPlane.getBlockAt(position);
+      if (ground.blockType === "water" || ground.blockType === "lava") {
+        this.groundPlane._data[this.groundPlane.coordinatesToIndex(position)] = new LevelBlock(blockType);
```
Hamms: use `setBlockAt` rather than `_data`

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -781,11 +781,19 @@ module.exports = class LevelModel {
         break;
     }
 
-    if (shouldPlace === true) {
-      var block = new LevelBlock(blockType);
+    if (entity === this.agent) {
+      let ground = this.groundPlane.getBlockAt(position);
+      if (ground.blockType === "water" || ground.blockType === "lava") {
```
Hamms: I propose we move as much of this logic into `Player` and `Agent` as we can; something like an `entity.canPlaceBlock` method that takes as arguments the blockType to be placed and the ground block it's being placed on top of. Then, `Player` can define the `wheat iff famland` logic and `Agent` can define the water/lava logic.

That'll be particularly nice, because we're going to want to expand the water/lava logic to make sure that only actual _blocks_ can be placed in water or lava, not things like redstone or wheat or rails.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -794,14 +794,14 @@ module.exports = class LevelView {
     return tween;
   }
 
-  playPlaceBlockAnimation(position, facing, blockType, blockTypeAtPosition, completionHandler) {
+  playPlaceBlockAnimation(position, facing, blockType, blockTypeAtPosition, entity, completionHandler) {
     var jumpAnimName;
     let blockIndex = this.yToIndex(position[1]) + position[0];
 
-    if (blockType === "cropWheat" || blockType === "torch" || blockType.startsWith("rail") || blockType.startsWith("redstoneWire")) {
+    if (entity === this.agent || blockType === "cropWheat" || blockType === "torch" || blockType.startsWith("rail") || blockType.startsWith("redstoneWire")) {
```
Hamms: similarly, let's move this blocktype checking into LevelBlock

src/js/game/GameController.js
```diff
@@ -1344,41 +1344,42 @@ class GameController {
   }
 
   placeBlock(commandQueueItem, blockType) {
-    const position = this.levelModel.player.position;
+    const player = this.getEntity(commandQueueItem.target);
```
Hamms: :+1:

src/js/game/LevelMVC/LevelModel.js
WilliamHarmonMS: Excellent point with placing redstone in water. I haven't even looked at what would happen with Redstone on the ground plane. I bet it is a hideous mess. Will address this on Monday.

src/js/game/LevelMVC/LevelModel.js
WilliamHarmonMS: I wanted to say there was some reason for this... Either that it was done that way before (in which case copy-paste and you're totally right) or that for some reason setBlockAt() caused a problem... I wanted to say setBlockAt() was being weird recently with something, but this might not be the PR. Will replace this logic and if anything funky happens I'll ping you.

src/js/game/LevelMVC/LevelView.js

src/js/game/LevelMVC/LevelModel.js
WilliamHarmonMS: Oh, it's because setBlockAt() does a number of things that presume it's the action plane. That's probably an erroneous presumption, so I'll look into refactoring that first 👍

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: So looking into this:
The playPlaceBlockAnimation doesn't -have- the block itself, just the blockType string. How do you want to proceed?
(1) keep it the way it is
(2) add a function to leveView that can evaluate blockType strings to determine whether or not the animation should play
(3) update every instance of this function call to take a block instead of a string (blockType) and then use a method in LevelBlock to do this check

src/js/game/LevelMVC/LevelModel.js
WilliamHarmonMS: Checked up on this this morning:
redstone Wire can now no longer be placed in water, but additionally, it cannot be placed below the Agent either. In actual minecraft "place below" will always fail if there is ground below you, even if the block (like wire or rails) doesn't take up the full index. There is no way to pace blocks within the index the Agent is in, so for consistency I have not allowed to place rails / redstone in the same slot.

src/js/game/LevelMVC/LevelView.js
Hamms: Of those three, I like (3) the best, but I'd also be fine with just adding a static method on to LevelBlock like we did with [`isMiniblock`](https://github.com/code-dot-org/craft/blob/master/src/js/game/LevelMVC/LevelBlock.js#L182)

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -803,31 +803,33 @@ module.exports = class LevelModel {
...
+      }
     }
 
     return placedBlock;
   }
 
+  alwaysPlaceActionPlane(blockType) {
+    return blockType.startsWith("redstone") || blockType.startsWith("rails") || blockType === "torch" || blockType.startsWith("piston");
+  }
+
```
Hamms: this is now unused

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -803,31 +803,33 @@ module.exports = class LevelModel {
...
-      placedBlock = this.actionPlane.setBlockAt(position, block);
-      this.player.isOnBlock = !block.isWalkable;
+      switch (result.plane) {
+        case "actionPlane":
+          placedBlock = this.actionPlane.setBlockAt(position, block);
+          entity.walkableCheck(block);
+          break;
+        case "groundPlane":
+          this.groundPlane.setBlockAt(position, block);
+          this.controller.levelView.refreshGroundPlane();
```
Hamms: why is it only necessary to refresh the ground plane, not the action plane?

src/js/game/Entities/Agent.js
```diff
@@ -19,6 +19,32 @@ module.exports = class Agent extends BaseEntity {
...
+    }
+    return result;
+  }
+
+  /**
+   * player walkable stuff
+   */
+  walkableCheck() {
+    //do nothing
+  }
```
Hamms: this should be defined on `BaseEntity`

WilliamHarmonMS: Fair enough. I was only thinking of the Agent since no other entity places blocks. This IS definitely safer though. So great point 👍

src/js/game/LevelMVC/LevelModel.js
WilliamHarmonMS: because the actionPlane isn't changing?

Hamms: I mean in the `"actionPlane"` case

WilliamHarmonMS: Oh wait, you totally mean why in this function. It's because when the actionPlane is changed, we don't refresh the whole thing. We only refresh a few local indices and that's done in setBlockAt()

We could totally add a general refreshGroundPlane to setBlokAt(). I'll do it that way :)

WilliamHarmonMS:
@Hamms PTAL 👍 

I added some test coverage in levelModeltest to make sure the placeblock() is still working like we expect, and the changes to levelPlane should be covered by the current tests actually, I believe. They didn't break anything and we're passing 62/62 tests, so hurray!

As always, test will continue to punch holes in this.

PR #127, opened 2017-09-21
Minecraft parity. Redstone charge can propagate a single actionBlock …

WilliamHarmonMS:
NOTE: powerCheckDeluxe is a not a real function name. This is to demonstrate functionality.

this.isSolid IS a real block attribute name because apparently that's literally how redstone charge propagation works in minecraft: if a block is transparent like Ice or Glass, it won't carry a charge. Though I'm up for changing it, it just might help inform future devs about whether or not a block should carry a charge?


WilliamHarmonMS:
Excellent ideas. Those will be quick fixes 👍

PR #128, opened 2017-09-21
Piston redstone

WilliamHarmonMS:
Pistons should destroy redstone and doors instead of pushing them. Apparently this functionality is needed for some of the early levels to work properly


src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -537,7 +550,7 @@ module.exports = class LevelPlane {
   getBlocksToPush(position, offsetX = 0, offsetY = 0) {
     let pushingBlocks = [];
     let workingPosition = position;
-    while (this.inBounds(workingPosition) && this.getBlockAt(workingPosition).blockType !== "") {
+    while (this.inBounds(workingPosition) && this.getBlockAt(workingPosition).blockType !== "" && !this.getBlockAt(workingPosition).blockType.startsWith("redstone") && !this.getBlockAt(workingPosition).blockType.startsWith("door")) {
```
Hamms: as a general rule, when logic about the identity of a block gets this complex it should really go in a method on LevelBlock itself, then we can just call `this.getBlockAt(workingPosition).getIsPushable()`

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1052,12 +1052,16 @@ module.exports = class LevelView {
         if (!this.controller.levelData.isEventLevel) {
           this.playPlayerAnimation("idle", playerPosition, facing, false, entity);
         }
-        this.playItemDropAnimation(destroyPosition, blockType, completionHandler);
+        if (completionHandler !== null) {
+          this.playItemDropAnimation(destroyPosition, blockType, completionHandler);
```
Hamms: I might be misunderstanding something about the completion handler, but don't we still want to play the animation?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -524,11 +526,22 @@ module.exports = class LevelPlane {
     let armBlock = new LevelBlock(pistonType);
     for (let i = blocksPositions.length - 1; i >= 0; --i) {
       let destination = [blocksPositions[i][0] + offsetX, blocksPositions[i][1] + offsetY];
+      let block = this.getBlockAt(blocksPositions[i]);
+      if (this.getBlockAt(destination) === undefined) {
+        console.log("break");
+      }
```
Hamms: nit: console.log

WilliamHarmonMS: I swear to god leaving these in my commits is my greatest weakness. I should just start eliminating the space between if and ( so that the linter won't even let me commit them.

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: 👍

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: When calling the animation where I was, there was no completion handler, so it would die 3-4 functions deeper when it tries to call completion handler animations. 

This is something I'll run by Cole / Joel. I know for doors getting destroyed, we don't even HAVE miniblocks for them to drop, but for redstoneWire, we totally do. If we're invested in making sure those still drop, this shouldn't be TOO bad to fix.

src/js/game/LevelMVC/LevelPlane.js
Hamms: Pro tip: the linter we use is actually hierarchical, so if you put your own `.eslint` file in a directory above the project directory, it'll run that one as well as the official one. So you can have your own custom eslint that enforces a superset of rules, like:

```javascript
module.exports = {
  "rules": {
    "no-console": ["error"]
  }
}
```

src/js/game/LevelMVC/LevelView.js
Hamms: What I mean is, we should probably add the `if (completionHandler)` check at the lowest level; the level where it is actually getting called

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -524,11 +526,19 @@ module.exports = class LevelPlane {
     let armBlock = new LevelBlock(pistonType);
     for (let i = blocksPositions.length - 1; i >= 0; --i) {
       let destination = [blocksPositions[i][0] + offsetX, blocksPositions[i][1] + offsetY];
+      let block = this.getBlockAt(blocksPositions[i]);
+      if (this.getBlockAt(destination).blockType.startsWith("redstone") || this.getBlockAt(destination).blockType.startsWith("door")) {
```
Hamms: might as well go ahead and _use_ `getIsPushable` here :)

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1104,7 +1106,9 @@ module.exports = class LevelView {
           (this.player.inventory[blockType] || 0) + 1;
         sprite.kill();
         this.toDestroy.push(sprite);
-        completionHandler();
+        if (completionHandler !== null) {
```
Hamms: all three of these `null` checks should just be `if (completionHandler)`; otherwise other values (for example, `undefined`) won't get handled correctly

WilliamHarmonMS: Roger that 👍

src/js/game/LevelMVC/LevelPlane.js
Hamms: In that case, we'll want a method for something like "getIsDestroyedWhenPushed" (which getIsPushable should probably use).

In general, we don't want stuff outside of LevelBlock to be making decisions by doing string comparisons on `blockType`; LevelBlock should be the only place that kind of logic lives

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1084,8 +1084,10 @@ module.exports = class LevelView {
       }
     });
     this.playScaledSpeed(explodeAnim.animations, "explode");
-    if (this.controller.getIsDirectPlayerControl() ^ !placeBlock) {
-      completionHandler();
+    if (this.controller.levelData.isEventLevel ^ !placeBlock) {
```
Hamms: this should still be using `getIsDirectPlayerControl`

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1095,7 +1097,7 @@ module.exports = class LevelView {
       sprite.sortOrder = this.yToIndex(destroyPosition[1]) + 2;
     }
 
-    if (this.controller.getIsDirectPlayerControl()) {
+    if (completionHandler && this.controller.getIsDirectPlayerControl()) {
```
Hamms: be careful about addition conditions to `if`s with an `else`; I believe in this scenario we want to avoid the `else` case when we're in direct control mode, so the `if (completionHandler)` should just go inside the block

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: 👍

PR #130, opened 2017-09-23
add getIsDirectPlayerControl helper to replace isEventLevel for character movement logic

Hamms:
This is necessary to enable the `isAgentLevel` setting to distinguish from `isEventLevel`


WilliamHarmonMS:
Darn linter 😂

Hamms:
Looks like it was actually a mock that needed updating

PR #146, opened 2017-09-28
'use' to hop on to a rail and ride around the track. 

WilliamHarmonMS:
@Hamms PTAL 👍 

This is a first pass at in game rails use. 'Use' on the rail and it'll make you go in the direction you're facing until you reach the end of the track. Normal movement disabled while on tracks. Once you stop, you ca move normally to leave the minecart.


src/js/game/Entities/Player.js
```diff
@@ -31,7 +32,7 @@ module.exports = class Player extends BaseEntity {
     const timeEllapsed = (+new Date() - this.lastMovement);
     const movementAlmostFinished = timeEllapsed > 300;
 
-    if ((queueIsEmpty || (queueHasOne && movementAlmostFinished)) && isMoving) {
+    if (!this.onTracks && ((queueIsEmpty || (queueHasOne && movementAlmostFinished)) && isMoving)) {
```
Hamms: can we pull this out into a "canUpdateMovement" method?

src/js/game/GameController.js
```diff
@@ -1173,6 +1173,7 @@ class GameController {
     let frontBlock = this.levelModel.actionPlane.getBlockAt(frontPosition);
 
     const isFrontBlockDoor = frontBlock === undefined ? false : frontBlock.blockType === "door";
+    const isFrontBlockRails = frontBlock === undefined ? false : frontBlock.blockType.startsWith("rails");
```
Hamms: shouldn't this be `frontBlock.isRail`, particularly since `railsRedstoneTorch` exists as a block type

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -49,6 +49,9 @@ module.exports = class LevelModel {
       if (this.actionPlane.getBlockAt(position).blockType.substring(0,12) === "redstoneWire") {
         this.actionPlane.determineRedstoneSprite(position);
       }
+      if (this.actionPlane.getBlockAt(position).blockType.startsWith("rails") && this.actionPlane.getBlockAt(position).blockType !== "railsRedstoneTorch") {
```
Hamms: similarly here, let's just use `.isRail`. Keep all the knowledge about what the blockType strings mean inside of LevelBlock

src/js/game/GameController.js
WilliamHarmonMS: Yeah we can change that. I was merely mimicking the "isFrontBlockDoor" definition there.

src/js/game/Entities/Player.js
WilliamHarmonMS: 👍

WilliamHarmonMS:
@Hamms  PTAL 👍

WilliamHarmonMS:
Note: This fails some unit tests, but the unit tests it's failing are also failing in master, so this PR is not actually introducing any failures. 👍

WilliamHarmonMS:
@Hamms PTAL whenever you have time. FINALLY FIX IT

src/js/game/GameController.js
```diff
@@ -1238,6 +1238,10 @@ class GameController {
           commandQueueItem.succeeded();
         });
       });
+    } else if (frontBlock.isRail) {
+      player.onTracks = true;
```
Hamms: any reason why this isn't inside `playTrack`?

src/js/game/LevelMVC/LevelView.js
```diff
@@ -608,14 +608,15 @@ module.exports = class LevelView {
...
       this.activateUnpoweredRails(unpoweredRails);
-      this.playTrack(position, this.player.facing, isOnBlock, completionHandler);
+      this.playTrack(position, this.player.facing, isOnBlock, this.player, completionHandler);
     });
   }
 
-  playTrack(position, facing, isOnBlock, completionHandler) {
-    const track = this.controller.levelModel.actionPlane.getMinecartTrack(position, facing);
+  playTrack(position, facing, isOnBlock, entity = this.player, completionHandler) {
+    let  track = this.controller.levelModel.actionPlane.getMinecartTrack(position, facing);
```
Hamms: any reason for this change?

src/js/game/GameController.js
WilliamHarmonMS: Because playTrack is called multiple times, and we only need to set it once when we hop on and once when we end.

It certainly COULD be in playTrack but we'd either do a bunch of setting to true when already true, or have an extra if statement to check if it's not already true?

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: Because when doing the redstone Rail work I change track to null so that the loop ends and the user stops when hitting uncharged rails. That's just a holdover from when I was ripping charged rail stuff out to get that darn integration test working

PR #153, opened 2017-09-29
request from design/art to add all the assets we had, but never hooke…

WilliamHarmonMS:
…d up


WilliamHarmonMS:
@Hamms @joshlory PTAL if this is good? Pretty straight forward. We should be able to merge this in with no complaints?

PR #159, opened 2017-09-29
fixing a game breaking redstone torch + piston bug

WilliamHarmonMS:


WilliamHarmonMS:
@Hamms PTAL 👍 

I was told this bug is game breaking and to prioritize it over anything else until its fixed because a few levels will be un-test-able next week if it's not addressed.

Thanks a ton!

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -602,8 +610,15 @@ module.exports = class LevelPlane {
     for (let i = blocksPositions.length - 1; i >= 0; --i) {
       let destination = [blocksPositions[i][0] + offsetX, blocksPositions[i][1] + offsetY];
       let block = this.getBlockAt(blocksPositions[i]);
-      if (this.inBounds(destination) && this.getBlockAt(destination).isDestroyableUponPush()) {
-        this.levelModel.controller.levelView.playExplosionAnimation(destination, 2, destination, block.blockType, null, null, this.player);
+      if (block.isDestroyableUponPush() || (this.inBounds(destination) && this.getBlockAt(destination).isDestroyableUponPush())) {
+        let newBlock = undefined;
+        if (block.isDestroyableUponPush()) {
```
Hamms: This seems a bit redundant. Why not

```javascript
if (block.isDestroyableUponPush()) {
  // destroy block
} else if (this.inBounds(destination) && this.getBlockAt(destination).isDestroyableUponPush()) {
  // destroy destination
}
```

WilliamHarmonMS: yeah... it's not pretty. For some reason, redstoneTorches were causing stack overflows and redstoneWire weren't. Something to do with being a power source? I'm not sure. 

the first block. check is for the torch, the second destination check is for the wire. I can try and clean it up, but right now it -is- working without crashing the game, which was what was happening in a number of the levels 😱

WilliamHarmonMS:
@Hamms PTAL <3

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -602,7 +610,7 @@ module.exports = class LevelPlane {
     for (let i = blocksPositions.length - 1; i >= 0; --i) {
       let destination = [blocksPositions[i][0] + offsetX, blocksPositions[i][1] + offsetY];
       let block = this.getBlockAt(blocksPositions[i]);
-      if (this.inBounds(destination) && this.getBlockAt(destination).isDestroyableUponPush()) {
+      if ((this.inBounds(destination) && this.getBlockAt(destination).isDestroyableUponPush())) {
```
Hamms: nit: unnecessary parens

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -683,6 +691,29 @@ module.exports = class LevelPlane {
...
+            case "U":
+              ignoreThisSide = [0, -1];
+              break;
+            case "L":
+              ignoreThisSide = [-1, 0];
+              break;
+            case "R":
+              ignoreThisSide = [1, 0];
+              break;
+          }
```
Hamms: This switch statement is very nearly repeating [this switch statement](https://github.com/code-dot-org/craft/blob/master/src/js/game/LevelMVC/LevelPlane.js#L555-L568), and is also relying on some subtle properties of the blockType string; both of these instances should be moved into LevelBlock

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -484,6 +491,7 @@ module.exports = class LevelPlane {
       this.setBlockAt(position, onPiston);
       this.pushBlocks(blocksPositions, offset[0], offset[1]);
     } else if (workingNeighbor.blockType === "") {
+      // Nothing to push, so just make the arm.
```
Hamms: :+1:

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: My goodness. I did this PR in such a rush as I was heading out of the office. That's embarrassing 😂

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: Yeah, good idea. That was a copy paste for functionality so I could get out of the office on time. Will move that over right now

src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -199,6 +199,22 @@ module.exports = class LevelBlock {
...
+      let direction = this.blockType.substring(6, 7);
+      switch (direction) {
+        case "D":
+          return "South";
+        case "U":
+          return "North";
+        case "L":
+          return "West";
+        case "R":
+          return "East";
```
Hamms: Could this use the FacingDirection constants instead?

joshlory:
Can you add a regression test to verify the fix?  Thanks!

joshlory:
I'm still seeing `Uncaught RangeError: Maximum call stack size exceeded` when stepping onto the east pressure plate in this sample level:

image

src/js/game/LevelMVC/LevelBlock.js
WilliamHarmonMS: They could? The numbers wouldn't be used for anything other than IDing the direction, (like the fact that West === 3 in terms of facing direction wouldn't come into play at all) but we could do that if you want?

WilliamHarmonMS:
Good catch, Josh! Just added pressurePlates to the levelBlock.isDestroyableUponPush() method. Should be good to go.

src/js/game/LevelMVC/LevelBlock.js
Hamms: Yeah, I mean for IDing the direction (to be clear, both here and in LevelPlane); since we already have a canonical way of identifying a direction, I'd rather use that than these arbitrary strings

WilliamHarmonMS:
@joshlory @Hamms PTAL whoever wants to. I know this is approved, but since I added the pressurePlate thing maybe someone wanted to take another look before hitting the nice big merge button.

WilliamHarmonMS:
This this was approved and all I've primarily added are unit tests that are passing I'm going to merge it up. 👍

PR #162, opened 2017-10-02
for level 1, we need to be able to spawn an agent on solutionCheck()

WilliamHarmonMS:
doAgentSpawn() is probably a bit messy. If the level is an agent level, doAgentSpawn() is called at init with a passed in levelData so you get the normal behavior. If you call it manually on a non-agent level, with a first argument of null, you spawn the agent at the index you passed in.

I've got ideas about how to clean this up, but also welcome any input you guys might want.

### DIFF
--- a/src/js/game/LevelMVC/LevelModel.js
+++ b/src/js/game/LevelMVC/LevelModel.js
@@ -68,7 +68,7 @@ module.exports = class LevelModel {
       this.controller.player = this.player;
 
       if (levelData.useAgent) {
-        this.doAgentSpawn(levelData);
+        this.spawnAgent(levelData);
       }
     }
 
@@ -76,13 +76,13 @@ module.exports = class LevelModel {
     this.computeFowPlane();
   }
 
-  doAgentSpawn(levelData = null, aX = -1, aY = -1) {
+  spawnAgent(levelData, positionOverride, directionOverride) {
     this.usingAgent = true;
     let [x, y] = [0, 0];
     let direction = 0;
     let name = "";
     let key = "";
-    if (levelData !== null) {
+    if (levelData) {
       [x, y] = levelData.agentStartPosition;
       direction = levelData.agentStartDirection;
       name = "PlayerAgent";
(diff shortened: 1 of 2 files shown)


src/js/game/LevelMVC/LevelModel.js
```diff
@@ -68,19 +68,44 @@ module.exports = class LevelModel {
       this.controller.player = this.player;
 
       if (levelData.useAgent) {
-        this.usingAgent = levelData.useAgent;
-        [x, y] = levelData.agentStartPosition;
-        const startingBlock = this.actionPlane.getBlockAt(levelData.agentStartPosition);
-        this.agent = new Agent(this.controller, "PlayerAgent", x, y, "Agent", !startingBlock.getIsEmptyOrEntity(), levelData.agentStartDirection);
-        this.controller.levelEntity.pushEntity(this.agent);
-        this.controller.agent = this.agent;
+        this.doAgentSpawn(levelData);
```
Hamms: nit; `spawnAgent` is a _lot_ more readable

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -68,19 +68,44 @@ module.exports = class LevelModel {
...
       }
     }
 
     this.computeShadingPlane();
     this.computeFowPlane();
   }
 
+  doAgentSpawn(levelData = null, aX = -1, aY = -1) {
```
Hamms: something like `agentSpawn(levelData, positionOverride, directionOverride)` would be a lot cleaner. It combines the X and Y into one position argument, which is semantically more accurate, adds the direction so we don't have to hardcode it, and also means that all arguments will default to `undefined` if not specified, which is again semantically more accurate than `null` or `-1`

WilliamHarmonMS: Oh right, because you don't need to give explicitly defined default arguments in JS. 😰 

I can change that for sure. I'm just in general not in love with having to add a case to the function that is -only- used for one special unique level but... I can make this cleaner.

WilliamHarmonMS:
@Hamms PTAL whenever you like. I addressed some of the PR feedback and hopefully it's less of a mess.

PR #168, opened 2017-10-02
decouple agent movement from WASD

WilliamHarmonMS:
Preventing WASD from controlling Agent


WilliamHarmonMS:
@Hamms @joshlory PTAL ❤️

Issue #81, opened 2017-09-06
Destroying Rail does not cause it to drop to the ground


v-samtha:
Found at: https://code-dot-org.github.io/craft/

Destroying a Rail does not cause a sprite of it to drop to the ground and be picked up by the Agent.

Steps:
1. Place Rails
2. Destory Rails
3. Observe item drop (or lack thereof)

Attachment:
[HoC_DestoryRail_RailIsNotDropped.zip](https://github.com/code-dot-org/craft/files/1283008/HoC_DestoryRail_RailIsNotDropped.zip)

joshlory:
Thanks for the bug report!  I think we'll need miniblock assets in order to implement the rail drop.

PR #173, opened 2017-10-02
look at all these hot new miniblocks

WilliamHarmonMS:
Resolves https://github.com/code-dot-org/craft/issues/81.


WilliamHarmonMS:
@Hamms @joshlory PTAL another super simple PR that'll get the destruction of our new blocks working like we want: No crashes, Yes mini-blocks 👍

test/unit/LevelBlockTest.js
```diff
@@ -27,7 +27,7 @@ test('getMiniblockFrame', t => {
   });
 
   // Rails give us nothing
-  t.equal(LevelBlock.getMiniblockFrame("rails"), undefined);
+  t.equal(LevelBlock.getMiniblockFrame("rails"), "railNormal");
```
joshlory: Nice!

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1111,8 +1165,16 @@ module.exports = class LevelView {
     }
   }
 
+  // Ice and glass blocks do not drop mini-blocks in Minecraft.
+  shouldDropMiniBlock(blockType) {
+    return !blockType.startsWith("ice") && !blockType.startsWith("glass");
```
joshlory: Can we add test coverage for these two cases as well?

WilliamHarmonMS: Sure thing!

WilliamHarmonMS: Added unit tests for both of these and they're passing, so PTAL? This should be clear to go I think.

Hamms: This seems pretty redundant. Given that `this.createMiniBlock` will simply return `null` for these values, we can just use `if (sprite)` to determine if we should be dropping a mini block.

I'll also point out that the return from `this.createMiniBlock` is determined by the definition of `LevelBlock.getMiniblockFrame`, which is already doing the work of specifying that ice and glass do not drop mini blocks. Redefining that here is definitely not great

WilliamHarmonMS: Removed and added the check in createMiniBlock()

WilliamHarmonMS:
@Hamms PTAL?

src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -229,9 +229,16 @@ module.exports = class LevelBlock {
...
+      return "redstoneTorch";
+    }
+
     if (blockType.startsWith("rails")) {
-      return;
+      return "railNormal";
+    }
+
+    if (blockType.startsWith("glass") || blockType.startsWith("ice")) {
+      return null;
```
Hamms: nit: I believe we want this to be `undefined`, not `null`

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1656,6 +1711,10 @@ module.exports = class LevelView {
     let sprite = null,
       frameList;
 
+    if (!LevelBlock.getMiniblockFrame(blockType)) {
+      return null;
+    }
+
```
Hamms: looks like this PR is actually made against a slightly old version of master; the latest version [already has an equivalent check](https://github.com/code-dot-org/craft/blob/master/src/js/game/LevelMVC/LevelView.js#L1714-L1717), making this redundant

WilliamHarmonMS:
Cool! I removed the redundant check and I changed the return time in getMiniblockFrame() for glass and ice to be undefined. Thanks a bunch @Hamms

PR #176, opened 2017-10-03
nether portals exist! after a fashion

WilliamHarmonMS:


WilliamHarmonMS:
@Hamms @joshlory Alright, fixed the unit test failure. Just a conflict resolution goof. This should be good to go.

src/js/game/LevelMVC/AssetLoader.js
```diff
@@ -105,6 +105,11 @@ module.exports = class AssetLoader {
         pngPath: `${this.assetRoot}images/Leaves_Spruce_Snowy_Decay.png`,
         jsonPath: `${this.assetRoot}images/Leaves_Spruce_Snowy_Decay.json`
       },
+      Nether_Portal: {
+        type: 'atlasJSON',
+        pngPath: `${this.assetRoot}images/Nether_Portal.png`,
+        jsonPath: `${this.assetRoot}images/Nether_Portal.json`
```
joshlory: Hmm, why are we adding the sprites to `Blocks.png` and also adding a `Nether_Portal.png`?

WilliamHarmonMS: Ah, missed that when taking out Nether_Portal.png

The phaser.sprite.animation frame stuff was much more ready to just pull from blocks.png (like it does for lava and water) than it was for adding a new file destination, for whatever reason. That can be taken out 👍

WilliamHarmonMS:
@joshlory PTAL good catch. I think this should be good now?

joshlory:
How is the portal defined in the level definition?

WilliamHarmonMS:
@joshlory The Nether_Portal is a bit goofy because it relies on specifically placing other blocks in other planes.

as the only block/structure that exists as a single entity that takes up multiple indices on the walk-able plane, the structure is as follows:

4 obsidian blocks to be placed on the groundPlane as the portals base (as this is not included in the animated sprite:
[obsidian][obsidian][obsidian][obsidian]
2 obsidian blocks to be placed on the actionPlane to serve as the sides of the portal to bother provide collision and so that shadows draw appropriately:
[obsidian][][][obsidian]
1 'Nether_Portal' block in the decoration layer in the same index as the left most actionPlane obsidian / groundPlane obsidian
[Nether_Portal][][][]

This is a super specific structure to level... 11 and users will never be able to destroy or create them. So long as the designers set it up right it should look A+

PR #182, opened 2017-10-03
adding audio assets for pressure plates, pistons, and the portal

WilliamHarmonMS:
@Hamms @joshlory PTAL

Just some audio assets from Minecraft for the pressure plate / pistons / nether portal


joshlory:
I think we'll want the mp3 version of these sounds as well — see https://github.com/code-dot-org/craft/commit/01dcd789853a694c6c4a07254eeb1961505dfa2a for a simple script to convert them.  Thank you!

WilliamHarmonMS:
@joshlory Per latest commit, new audio assets with advanced mp3 tech 👽

PR #193, opened 2017-10-04
URGENT FIX: crash on refreshActionPlane + diamondMiniblock

WilliamHarmonMS:
This is causing crashes when players try and get diamond blocks.

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -1420,7 +1420,7 @@ module.exports = class LevelView {
           return;
         }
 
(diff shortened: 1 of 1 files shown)


src/js/game/LevelMVC/LevelView.js
```diff
@@ -1420,6 +1420,10 @@ module.exports = class LevelView {
           return;
         }
 
+        if (newBlock && newBlock.blockType === "diamondMiniblock") {
```
Hamms: this should instead be `if (newBlock && newBlock.getIsMiniblock())`

joshlory:
Test coverage?

PR #196, opened 2017-10-04
Add some tweaks to diamond progression path

Hamms:
We have slightly different requirements for Diamond miniblocks than we do for the standard miniblocks dropped when destroying other blocks. Namely, we want the collectible range to be quite a bit smaller, so you have to actually land on the diamond's square to collect it, and we also want the randomized drop location to be a bit less variable, so it's guaranteed to always be visually obvious which square the diamond is actually on.

This PR represents one such approach to implementing those requirements. Another way to do it would be to override miniblock behavior for all of agent, so that the miniblocks defined at level-creation time behave exactly the same as standard miniblocks. This would have the advantage of a more consistent user experience, but I think that having these diamonds be _inconsistent_ is actually kind of desirable. I'm definitely open to thoughts on that.


Hamms:
It's ready!

Hamms:
Will fix https://github.com/code-dot-org/craft/pull/196

PR #197, opened 2017-10-04
fixed a problem where pressure plates would be reset if the agent and…

WilliamHarmonMS:
… player shared an index

Resolves https://github.com/code-dot-org/craft/issues/206.

### DIFF
--- a/src/js/game/Entities/Agent.js
+++ b/src/js/game/Entities/Agent.js
@@ -213,4 +213,9 @@ module.exports = class Agent extends BaseEntity {
         // there is no entity
         && (frontEntity === undefined);
   }
+
+  canTriggerPressurePlates() {
+    return true;
+  }
+
 };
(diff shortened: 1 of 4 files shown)


WilliamHarmonMS:
Well we don't want -any- entity to trigger a pressure plate. Only the player or Agent. If the player is moving off a pressure plate the only thing he needs to be concerned with is if the Agent is still there? I mean you're right it's not referencing the plate exactly, but that is the only case that should leave the pressure plate triggered after moving off it.

Though we could still say "getblockat(position) === pressure plate && agent.position === position"

or something like that?

Hamms:
Let's then add an Entity method `canTriggerPressurePlate`, and check for entities that fulfill that requirement

WilliamHarmonMS:
@Hamms PTAL 👍

joshlory:
Test coverage?

WilliamHarmonMS:
Given that this was approved, I'm going to merge it through and will make another PR for some unit tests 👍 

Just want to make sure it's included for tomorrows tests if possible

PR #201, opened 2017-10-04
agent can now move through Steve

WilliamHarmonMS:
Agent needs to not be blocked by steve, just like steve isn't blocked by the agent.

Resolves https://github.com/code-dot-org/craft/issues/205.

### DIFF
--- a/src/js/game/Entities/Agent.js
+++ b/src/js/game/Entities/Agent.js
@@ -206,15 +206,7 @@ module.exports = class Agent extends BaseEntity {
     }
   }
 
-  hasPermissionToWalk(actionBlock, frontEntity) {
-        return (actionBlock.isWalkable || ((frontEntity !== undefined && frontEntity.isOnBlock)
-        // action plane is empty
-        && !actionBlock.isEmpty))
-        // there is no entity
-        && (frontEntity === undefined || this.entityIsPlayer(frontEntity));
-  }
-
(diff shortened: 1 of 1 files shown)


WilliamHarmonMS:
Well, canMoveThrough() for Agent is always true. But that's not the case for all entities. Cows and sheeps and stuff would not move through the Agent. In minecraft they move around the Agent (though they don't prevent the Agent from moving, he would likely just "bump" them out of the way, which is something that we could do at some point but not something I'm super worried about). The only thing that can actively move through the Agent with no repercussions is the player.

like wise, the Agent moving through the player is without repercussion, but as I said he bumps other entities aside a bit (he moves through regardless, but they make a token effort to get out of the way).

I'm a bit wary about using "targetEntity.canMoveThrough() || this.canMoveThrough()" as a part of the check because of how systemic that permission would be.

Hamms:
Ahh, that makes sense; I didn't realize that the agents move through property was unique to the player. This means that we'll likely need to update the existing move through logic, since right now I think anything will be able to move through it.

This LGTM for now

joshlory:
Until non-player entities have logic to move around the Agent, should we let them move through the Agent?  What is the downside?

WilliamHarmonMS:
>"Until non-player entities have logic to move around the Agent"

Do the other entities not do A*? I swear I saw an A* logic somewhere.

PR #207, opened 2017-10-04
fixed a piston arm retraction bug in very niche cases, but was causin…

WilliamHarmonMS:
…g problems on level 10


WilliamHarmonMS:
@Hamms PTAL

Issue #141, opened 2017-09-27
Piston head shadowing is on top of pushed blocks


v-samtha:
Found at: https://code-dot-org.github.io/craft/

Giving power to a Piston causes the head of the piston to extend. However, after pushing a block (or coming into contact with a block after being extended), the shadowing on the Piston head is on top of the block giving it the impression that the piston is over the block not flush against it.

Steps:
1. Place a Piston
2. Place a moveable block at the head of the Piston
3. Hook up Redstone to the Piston
4. Power the Piston
5. Observe Piston head shadowing

Attachments:
![hoc_piston_extended_headshadowing](https://user-images.githubusercontent.com/31709758/30938573-50606c12-a38f-11e7-849f-162551270a5b.png)

joshlory:
@WilliamHarmonMS I think this is just a sprite sort order fix?

WilliamHarmonMS:
@joshlory you got it, captain. We just need piston arms to draw like... 0.1f further back on the z-plane so that they're still on top of everything they should be (like the ground plane and all that) but that they're behind other actionPlane objects on the same row. I've been meaning to do that for a while, but just haven't gotten around to it because it doesn't actually effect functionality at all 😫

PR #219, opened 2017-10-05
piston arm shadows no longer overlay over actionPlane blocks

WilliamHarmonMS:
Prevents piston arm shadows from drawing on top of actionPlane blocks.


WilliamHarmonMS:
@joshlory PTAL

fix for [issue#141](https://github.com/code-dot-org/craft/issues/141)

Issue #184, opened 2017-10-03
Wooden Door open and close animation may not complete


v-samtha:
Found at: https://levelbuilder-studio.code.org/s/colehoc2017/stage/1/puzzle/1

A player sitting in front of an open door opens and closes it by using the Use button. During some of these animation cycles, the door may not completely open (or close) leaving the door at a weird angle. Note, the state of the door is correct, as player can walk through (if open) or be blocked (if closed). It is just visually, the door looks ajar.

Steps:
1. Load any level with a Wooden Door (e.g. Level 1)
2. Select "Run"
3. Walk up to a Wooden Door
4. Open and close the Wooden Door repeatedly
5. Observe animations

Attachments:
[HoC_WoodenDoor_Animation_MayNotOpenOrCloseAllTheWay.zip](https://github.com/code-dot-org/craft/files/1353824/HoC_WoodenDoor_Animation_MayNotOpenOrCloseAllTheWay.zip)

PR #225, opened 2017-10-05
Fix stuck door animation

joshlory:
I _think_ this should resolve https://github.com/code-dot-org/craft/issues/184.

[Phaser 2.3.0 had a fix for this exact issue](https://github.com/photonstorm/phaser/issues/1628#issuecomment-334590492), but I think the fix may be incomplete.  I wasn't able to repro the issue after adding a call to `animation.updateCurrentFrame();` in the onAnimationEnd callback.

This change also updates the animation to change speed with the slider.  This won't affect normal gameplay, but ensures that speeding up or slowing down game speed will also modify the door animation speed.


src/js/game/LevelMVC/LevelView.js
```diff
@@ -467,7 +467,9 @@ module.exports = class LevelView {
     let blockIndex = (this.yToIndex(position[1])) + position[0];
     let block = this.actionPlaneBlocks[blockIndex];
     let animationName = open ? "open" : "close";
-    this.onAnimationEnd(block.animations.play(animationName, 60), () => {
+    const animation = this.playScaledSpeed(block.animations, animationName);
+    this.onAnimationEnd(animation, () => {
+      animation.updateCurrentFrame();
```
Hamms: is this (calling `updateCurrentFrame` on animation end) something we should be doing in general?

joshlory: TBH I'm not sure.  It's an undocumented API...  Until this specific issue is fixed upstream I think we need this to fix the door animation.  Other animations loop or disappear when complete.

I'm tempted to only patch this here, then revisit if we see another case of this bug with a different animation.  What do you think?

joshlory: Also not all animations are connected to an `onAnimationEnd` event.  Is there one place we can add this to fix this class of issues?

Hamms: There unfortunately is not one place we could go to fix this kind of issue, but it might be nice at some point to add one so our animations can in general be a little more standardized.

I think fixing this just here makes sense for now

PR #227, opened 2017-10-05
De-duplicate controller code for choosing a target entity

joshlory:
Found via the [code coverage report for GameController.js](https://codecov.io/gh/code-dot-org/craft/src/436c74beb4afa0f653154a41cb6d44d081c7f38a/src/js/game/GameController.js). There's more cleanup to do, but starting here for now 😄.

### DIFF
--- a/src/js/game/GameController.js
+++ b/src/js/game/GameController.js
@@ -955,13 +955,6 @@ class GameController {
     }
   }
 
-  useEntity(commandQueueItem, userIdentifier, targetIdentifier) {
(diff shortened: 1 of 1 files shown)


src/js/game/GameController.js
```diff
@@ -1057,13 +955,6 @@ class GameController {
     }
   }
 
-  useEntity(commandQueueItem, userIdentifier, targetIdentifier) {
```
Hamms: :+1:

PR #235, opened 2017-10-06
Support blocks that believe they should drop a miniblock, but for which the level view has no mini block frame data

### LINKED ISSUE #137: [CRASH] Destroying a Snow block causes lockup/crash
Found at: https://code-dot-org.github.io/craft/

A player during gameplay comes across a Snow block and attempts to break break it. This causes the cracking animation and the particle explosion effect to occur. However, in the middle of the explosion affect, the game appears to lock up (possibly crash?).

Note, due to the white nature of a bunch of blocks (snow, quartz, etc.) we are not 100% certain that this is a Snow block. However, due to the snow biome it is in, we are making this assumption.

Steps:
1. Place Snow block
2. Destory the placed Rail
3. Observe lockup/crash

Attachments:
None. Map updated before video and callstack could be grabbed. Should be similar to Rail issue in #126.

Hamms:
Fixes #137

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -1792,7 +1792,7 @@ module.exports = class LevelView {
     const yOffsetRange = overrides.yOffsetRange || 40;
 
     const frame = LevelBlock.getMiniblockFrame(blockType);
-    if (!frame) {
+    if (!(frame && this.miniBlocks[frame])) {
       return sprite;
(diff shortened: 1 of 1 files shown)


WilliamHarmonMS:
Good contingency, but all destroyable blocks should have mini-blocks now so HOPEFULLY this wouldn't be an issue. But great future proofing because I'm sure future projects will add stuff.

PR #237, opened 2017-10-06
Add redstone torch, pressure plate & pistons to the playground

joshlory:
Make it possible to place more blocks in the playground from the dropdown menu.


joshlory:
The page that loads when you run `grunt`.  Also mirrored here: https://code-dot-org.github.io/craft/

Issue #118, opened 2017-09-19
Doors close if Steve/Agent are in doorway when unpowered


v-samtha:
Found at: https://code-dot-org.github.io/craft/

Steve/Agent is standing in a doorway of a powered Iron Door. While there, the Iron Door becomes unpowered and attempts to close. Instead of remaining open while an object is present, the door closes and causes the sprites to collide.

Steps:
1. Place Iron Door on ground
2. Power the Iron Door
3. Put Steve or Agent in the doorway
4. Unpower the Iron Door
5. Observe door state

Attachments:
[HoC_Agent_InDoorway_ClosesWhenUnpowered.zip](https://github.com/code-dot-org/craft/files/1316120/HoC_Agent_InDoorway_ClosesWhenUnpowered.zip)

PR #240, opened 2017-10-06
agent moving into door fix

WilliamHarmonMS:
@joshlory Not in love with a lot of this. The controller.moveForward gets a bit weird because the first half is calculate pre-model update, and then the controller.execute is called and we handle things after the model update.

Basically the GameController changes are to make sure the Agent's moveState is updated (and I knew putting the logic there achieved the desired functionality, though honestly it should probably live in doMoveForward() in Agent.js or something). This is important because it's used in the door animation exceptions.

Most of the LevelPlane.js changes are about getting around the weak charge checks returning false and closing the door on you.

This has sort of become somewhat of a mess, ESPECIALLY since it's for that one turbo niche case of door right above pressure plate and then a user walking between the two repeatedly.

However, this DOES all work and it'd be awesome to see it for Monday for the user test because I have a sneaking suspicion that they're changing some of the levels to have pressurePlates below doors. 😩

### DIFF
--- a/src/js/game/GameController.js
+++ b/src/js/game/GameController.js
@@ -794,22 +794,36 @@ class GameController {
 
   moveForward(commandQueueItem) {
     const target = commandQueueItem.target;
-    const moveOffset = this.directionToOffset(target.facing);
+    let entity = this.getEntity(target);
+    if (entity === this.agent) {
+      entity.movementState = entity.facing;
+    }
+    const moveOffset = this.directionToOffset(entity.facing);
     this.handleMoveOffPressurePlate(commandQueueItem, moveOffset);
 
     this.execute(commandQueueItem, 'moveForward');
 
+    if (entity === this.agent) {
+      entity.movementState = -1;
+    }
     this.handleMoveOnPressurePlate(commandQueueItem, moveOffset);
     this.handleMoveOffIronDoor(commandQueueItem, moveOffset);
   }
 
   moveBackward(commandQueueItem) {
     const target = commandQueueItem.target;
-    const moveOffset = this.directionToOffset(FacingDirection.opposite(target.facing));
+    let entity = this.getEntity(target);
+    if (entity === this.agent) {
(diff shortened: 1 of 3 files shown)


src/js/game/GameController.js
```diff
@@ -794,22 +794,36 @@ class GameController {
 
   moveForward(commandQueueItem) {
     const target = commandQueueItem.target;
-    const moveOffset = this.directionToOffset(target.facing);
+    let entity = this.getEntity(target);
+    if (entity === this.agent) {
+      entity.movementState = entity.facing;
```
joshlory: Why does this special logic only apply to the Agent?

joshlory: (Same question below.  In theory this should also work for a Player that is controlled by code, right?)

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -410,14 +479,18 @@ module.exports = class LevelPlane {
       if (block.blockType === "doorIron" && position !== positionInQuestion) {
         block.isPowered = this.powerCheck(position, true);
         if (block.isPowered && !block.isOpen) {
-          block.isOpen = true;
           if (this.levelModel) {
-            this.levelModel.controller.levelView.animateDoor(index, true);
+            if (!this.skipDoorFrom(position)) {
```
joshlory: I wonder if it would be simpler to have the door check if any entity is present at the current index.  If yes, add a listener to the entity (or entities) to check again on their next movement.  What do you think?

joshlory: There are multiple ways the door can try to close, beyond a player moving off a pressure plate.  Here's another example: https://github.com/code-dot-org/craft/issues/118.

WilliamHarmonMS:
KILLING IT BECAUSE NEW PR IS FAR SUPERIOR

Issue #216, opened 2017-10-05
Add integration tests for 2017 levels


joshlory:
Tracking issue.

joshlory:
@WilliamHarmonMS I think this is complete now 🎉?  I'm looking at https://github.com/code-dot-org/craft/blob/master/test/integration/AgentTest.js.

Want to close it out?

PR #243, opened 2017-10-09
Initial Agent level integration test

joshlory:
Add test fixture and integration test for the first Agent level.  The actual level will change, but this is a good baseline to make sure we don't regress pressure plate and Agent behavior.

### DIFF
--- a/test/integration/AgentTest.js
+++ b/test/integration/AgentTest.js
@@ -1,0 +1,24 @@
+const test = require("tape");
+const attempt = require("../helpers/RunLevel.js");
+
+test('Agent 1: Spawn Agent', t => {
+  attempt('agent01', api => new Promise(resolve => {
+    // Queue commands to move the player inside the house.
+    for (let i = 0; i < 2; i++) {
+      api.turnRight(null, 'Player');
+      for (let j = 0; j < 5; j++) {
+        api.moveForward(null, 'Player');
+      }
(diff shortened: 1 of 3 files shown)


test/integration/AgentTest.js
```diff
@@ -0,0 +1,24 @@
...
+    for (let i = 0; i < 2; i++) {
+      api.turnRight(null, 'Player');
+      for (let j = 0; j < 5; j++) {
+        api.moveForward(null, 'Player');
+      }
+    }
+
+    // Run the level. When finished, check that the Agent has spawned.
+    api.startAttempt((success, levelModel) => {
+      t.deepEqual(levelModel.player.position, [3, 3]);
```
jeffmckune95: Relying on agent position seems fragile.

joshlory: Good call.  I've updated this test to also verify that the Agent is not present at the start of the level, and spawns when the player completes the level.  Does this approach look ok?

joshlory: The hardcoded `[3, 3]` position is a bit fragile, but it's tied to a fixture that is checked in specifically for the integration tests.  The fixture won't change unless we intentionally update it to match any live-site level changes.

joshlory:
See issue https://github.com/code-dot-org/craft/issues/216.

PR #245, opened 2017-10-09
Add integration test for Agent level five

joshlory:
The actual level may change, but this should be a good baseline to make sure we don't regress Agent behavior.

(I'm skipping ahead a few levels here because their design is still in flux.)


joshlory:
See issue https://github.com/code-dot-org/craft/issues/216.

Issue #220, opened 2017-10-05
While standing on a pressure plate, bumping into an obstacle re-triggers the pressure plate


joshlory:
Is this a bug, or expected behavior?

![pressure-plate-reset](https://user-images.githubusercontent.com/413693/31242940-901dded2-a9bd-11e7-9383-5c191a61e144.gif)

WilliamHarmonMS:
I think it's a bug. Does it remain closed, or does it re-open? I think it's just because the door-close-because-leaving-index is done before we know if we bumped or not. 😩

PR #246, opened 2017-10-10
doors should no longer animate when bumping but not leaving pressureP…

WilliamHarmonMS:
…lates

Fixes [issue#220](https://github.com/code-dot-org/craft/issues/220)

### DIFF
--- a/src/js/game/Entities/BaseEntity.js
+++ b/src/js/game/Entities/BaseEntity.js
@@ -213,7 +213,12 @@ module.exports = class BaseEntity {
         let forwardPosition = this.controller.levelModel.getMoveForwardPosition(this);
         var forwardPositionInformation = this.controller.levelModel.canMoveForward(this);
         if (forwardPositionInformation[0]) {
+            let offset = this.controller.directionToOffset(this.facing);
+            let weMovedOnTo = this.handleMoveOnPressurePlate(offset);
             this.doMoveForward(commandQueueItem, forwardPosition);
+            if (!weMovedOnTo) {
+              this.handleMoveOffPressurePlate(this.reverseOffset(offset));
+            }
         } else {
             this.bump(commandQueueItem);
             this.callBumpEvents(forwardPositionInformation);
@@ -227,7 +232,12 @@ module.exports = class BaseEntity {
         let backwardPosition = this.controller.levelModel.getMoveBackwardPosition(this);
         var backwardPositionInformation = this.controller.levelModel.canMoveBackward(this);
         if (backwardPositionInformation[0]) {
+            let offset = this.controller.directionToOffset(FacingDirection.opposite(this.facing));
+            let weMovedOnTo = this.handleMoveOnPressurePlate(offset);
             this.doMoveBackward(commandQueueItem, backwardPosition);
+            if (!weMovedOnTo) {
+              this.handleMoveOffPressurePlate(this.reverseOffset(offset));
+            }
         } else {
             this.bump(commandQueueItem);
             this.callBumpEvents(backwardPositionInformation);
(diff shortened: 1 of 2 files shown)


WilliamHarmonMS:
BaseEntity, or any entity, never calls handleMoveOffPressurePlate currently. It's directly called in 3 cases in GameController (moveDirection, moveForward, and moveBackward) before the execute(CQI, 'command_type') call, where handleMoveOnPressurePlate / handleMoveOffIronDoor are then called. 

Basically, the way it's set up, is that handling the pressurePlate is evaluated before we even reach the entity specific functions. Given that handling the behavior of the pressurePlate is likely uniform across all entities (all entities that press it down by being on it, would also release it by moving off and all that), and that this concerns the state of non-entity objects (the plate and the door) I'm sort of ok with it being handled in the controller calls? I mean, I'd ideally move it to the LevelPlane if we were making the system from scratch (as the door is on the model's actionPlane, the plane should manage the state of its "blocks" and then communicate that to the view for animating). But I'm not sure redoing all that is an efficient use of time at this stage?

We could go about moving it all into player, but we'd have to change the way controller.handleMoveOffPressurePlate() and controller.moveOnPressurePlate() work to do so (I believe). Let me know what you think.

joshlory:
If we move it all into `Player.js` don't we have the same issue with the Agent?

joshlory:
My preference would be to do it the "right" way.  I'd be in favor of moving the logic to `BaseEntity.js` if it makes the code cleaner and reduces edge cases.  We still have enough time before shipping to implement refactors that make sense.

WilliamHarmonMS:
Yeah, my bad, I meant base entity. Pressure Plate management should definitely be a thing all entities are concerned with, so if we were going to move it into an entity function, that'd be the place to do it.

But at that point we should discuss how we want to structure the system. I'm fine calling an entity.handPlateStuff() in lieu of the controller.handlePlateStuff. I can take a pass at the refactor this morning, and we can go over it during stand up 👍

src/js/game/Entities/BaseEntity.js
```diff
@@ -597,4 +598,33 @@ module.exports = class BaseEntity {
         // no lava or water
         && (groundBlock.blockType !== "water" && groundBlock.blockType !== "lava");
   }
+
+  handleMoveOffPressurePlate(moveOffset) {
```
joshlory: This looks great!  Can we now call `handleMoveOffPressurePlate` and `handleMoveOnPressurePlate` from `BaseEntity.prototype.doMoveForward`?

WilliamHarmonMS: Would you like controller.execute to be refactored to pass along moveOffset? Since entity.handleMoveOffPressurePlate(moveOffset); takes it as an argument, and it never makes it to the entity.doMoveForward().

if we're going to pass along that data, it would have to make its way into controller.execute... and then into entity.addCommand()... I'm not sure the best way to retain that data without making this -really- overly arduous.

Will continue looking into it 👍

WilliamHarmonMS: Update:

The information is captured in baseEntity actually, but the one problem is that doMoveForward() behaves a bit differently depending on function call. 

Basically: If calling moveForward / moveBackward, when we hit doMoveForward() in baseEntity, movementState will be -1, and facing direction will be 0-3. So we'd use facing direction to get offset.

if we call moveDirection() however, facing direction will be wrong. It will be the way the entity WAS facing before the move call and movementState will be 0-3, so that is what we'd use to get offset.

problem is by the time we hit baseEntity in doMoveForward, I don't think we capture what command that call is coming from? This isn't a problem for the Agent because it can't call moveDirection, but the Player causes an issue... We can assume always moveDirection() for this year, but it wouldn't work correctly for past years... not that past years have pressure plates... but it feels bad.

src/js/game/GameController.js
```diff
@@ -715,7 +715,11 @@ class GameController {
   handleMoveOffPressurePlate(commandQueueItem, moveOffset) {
```
joshlory: Can this (and `GameController.prototype.handleMoveOnPressurePlate`) go away now?

src/js/game/Entities/BaseEntity.js
```diff
@@ -212,7 +213,12 @@ module.exports = class BaseEntity {
         let forwardPosition = this.controller.levelModel.getMoveForwardPosition(this);
         var forwardPositionInformation = this.controller.levelModel.canMoveForward(this);
         if (forwardPositionInformation[0]) {
+            let offset = this.controller.directionToOffset(this.facing);
+            let weMovedOnTo = this.handleMoveOnPressurePlate(offset);
             this.doMoveForward(commandQueueItem, forwardPosition);
+            if (!weMovedOnTo) {
+              this.handleMoveOffPressurePlate(this.reverseOffset(offset));
+            }
```
joshlory: Looks great 🎉

The only case I see this might not handle is multiple entities both standing on the pressure plate, then one steps off.  But that's such an edge case I don't think we're worried about it for now.  This is a big improvement!

WilliamHarmonMS: I actually think that's handled elsewhere (unless I botched something here).

WilliamHarmonMS:
PTAL <3

WilliamHarmonMS:
Roger that! Will make another test for some tests in..... levelEntity test probably?

PR #250, opened 2017-10-10
Make it possible to load test fixture levels in the Playground app

joshlory:
Sample:

screen shot 2017-10-10 at 3 22 46 pm


joshlory:
Sure thing!  Future PR?  Right now none of the Adventurer fixture levels are playable because they fail as soon as all the command have finished running (right now zero commands).

Issue #175, opened 2017-10-03
Agent incorrectly plays walking SFX when walking over water and lava


v-samtha:
Found at: https://levelbuilder-studio.code.org/s/colehoc2017/stage/1/puzzle/6

The Agent when stepping over water or lava makes a sound effect as if they are stepping on land. In Minecraft, the Agent only makes walking sound effects if they are walking across land. When on water or lava, the Agent hovers and thus make no sounds.

Steps:
1. Load any level with water or lava (e.g. Level 6)
2. Use code to make the Agent walk over water or lava
3. Listen to the SFX played

Attachments:
[HoC_Agent_WalkingOnWater_SFXIncorrectlyPlayed.zip](https://github.com/code-dot-org/craft/files/1350765/HoC_Agent_WalkingOnWater_SFXIncorrectlyPlayed.zip)

PR #251, opened 2017-10-10
the agent shouldn't make noise when walking over water

WilliamHarmonMS:
Fixes https://github.com/code-dot-org/craft/issues/175.

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -851,6 +851,9 @@ module.exports = class LevelView {
 
   playBlockSound(groundType) {
     var oreString = groundType.substring(0, 3);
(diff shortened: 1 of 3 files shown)


joshlory:
Also Lava?

src/js/game/LevelMVC/LevelView.js
```diff
@@ -851,6 +851,9 @@ module.exports = class LevelView {
 
   playBlockSound(groundType) {
     var oreString = groundType.substring(0, 3);
+    if (groundType === "water" || groundType === "lava") {
+      return;
+    }
```
joshlory: Tiny nit: I'd rather see this above `var oreString = groundType.substring(0, 3);` so that line still has context for how it is used.

Issue #221, opened 2017-10-05
Agent is always rendered on top of Steve


v-samtha:
Found at: https://levelbuilder-studio.code.org/s/colehoc2017/stage/1/puzzle/2

Whenever both Steve and the Agent occupy the same block, the Agent is always rendered on top of Steve. After talking with William, it seems that the current plan is to have Steve always render on top of the Agent. However, a new sprite sheet is coming for the Agent which may change this.

Thus, this bug is more for tracking purposes to ensure that when the final decision is made that it is properly implemented.

Steps:
1. Load Level 2 (or any subsequent level with the Agent
2. Write code to move the Agent onto of Steve
3. Select "Run"
4. Move Steve and the Agent on top of each other
5. Observe rendering order

Attachments:
[HoC_Agent_AlwaysRenderedOnTopOfSteve.zip](https://github.com/code-dot-org/craft/files/1360398/HoC_Agent_AlwaysRenderedOnTopOfSteve.zip)

Hamms:
The information I received from Joel was "I believe the Agent should draw on top, since it is smaller," so that's how I implemented it. Changing the implementation once we figure out for sure what we want it to do will just be a matter of updating [this method](https://github.com/code-dot-org/craft/blob/master/src/js/game/Entities/Agent.js#L48-L55) to return an offset of `- 1` rather than `+ 1`

alalb:
@Hamms - I spoke to Joel yesterday and decided that we want the **player** to be drawn on **top** of the agent all the time, and since we are making the agent bigger it will not look so good.
and the player will be able to easily move the player with the arrow keys.

joshlory:
Should be fixed with PR https://github.com/code-dot-org/craft/pull/247.

joshlory:
Note that there's a small regression with the z-ordering when walking over the Agent.  The new z-ordering makes the Agent look like a flat tile, and the Player walks over the Agent's head.

Before:

![player-walk-over-agent-good](https://user-images.githubusercontent.com/413693/31412124-b9554bdc-adc8-11e7-882d-1bc32d3bf394.gif)

After:

![player-walk-over-agent-bad](https://user-images.githubusercontent.com/413693/31412105-b49df03a-adc8-11e7-81b0-a42423a44617.gif)

PR #252, opened 2017-10-10
Update sortOrder Index during movement animation, rather than before

Hamms:
[First](https://github.com/code-dot-org/craft/commit/8b52ea2aceae19688bf2e4ef1649dfc34cf90908), I update the `playMoveForwardAnimation` and `playMoveBackwardAnimation` methods to significantly reduce the amount of repeated code between them.

[Second](https://github.com/code-dot-org/craft/commit/94292a7f0b5089e61d2ee57360623a88438c86b2), I update the walk animation to actually hook into the tween and update the entity sortOrder at (approximately) the halfway point of the animation, rather than always updating it at the beginning. This means that the nice walking overlap effect happens for both entities in both directions, rather than before where it would only happen for the entity being drawn on the "bottom"

Entity | Before | After
--- | --- | ---
Agent | agent-before | agent-after
Player | player-before | player-after

### DIFF
--- a/src/js/game/Entities/Agent.js
+++ b/src/js/game/Entities/Agent.js
@@ -146,7 +146,7 @@ module.exports = class Agent extends BaseEntity {
       groundType = levelModel.actionPlane.getBlockAt(player.position).blockType;
     }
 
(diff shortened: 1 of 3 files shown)


joshlory:
Related to this issue: https://github.com/code-dot-org/craft/issues/221#issuecomment-335616434.

PR #256, opened 2017-10-11
Add level select to the Playground

joshlory:
Follow up to PR https://github.com/code-dot-org/craft/pull/250.

image


joshlory:
Only the test-fixture version of the levels.  So it'll be the same as the static snapshot we use for integration tests.  But yes, a version of the new levels!

PR #266, opened 2017-10-11
updating movement state when moveForward() moveBackward()

WilliamHarmonMS:
@joshlory This will set the movestate to match the facing direction (or reverse, if move backwards) and then set it back to -1 once doMoveForward is executed (if this isn't reset, the command never stops and the entity will repeatedly call doMoveForward at every opportunity). 

This should allow for the bump refactor to be much cleaner.


WilliamHarmonMS:
I THINK we're going to scrap this whole PR provided PR#246 goes through.

PR #268, opened 2017-10-11
idle animation call no longer assumes the player is the only entity t…

WilliamHarmonMS:
…o ever exist

May add some more to this. Not ready for a look yet.


WilliamHarmonMS:
@joshlory  or @Hamms this should be ready to go. PTAL

PR #270, opened 2017-10-12
doors should no longer close on the agent

WilliamHarmonMS:
@joshlory that sure was some gratifying clean up.

### DIFF
--- a/src/js/game/Entities/BaseEntity.js
+++ b/src/js/game/Entities/BaseEntity.js
@@ -225,6 +225,7 @@ module.exports = class BaseEntity {
             if (!weMovedOnTo) {
               this.handleMoveOffPressurePlate(this.reverseOffset(offset));
             }
+            this.handleMoveOffIronDoor(this.reverseOffset(offset));
         } else {
             this.bump(commandQueueItem);
             this.callBumpEvents(forwardPositionInformation);
@@ -244,6 +245,7 @@ module.exports = class BaseEntity {
             if (!weMovedOnTo) {
               this.handleMoveOffPressurePlate(this.reverseOffset(offset));
             }
+            this.handleMoveOffIronDoor(this.reverseOffset(offset));
         } else {
             this.bump(commandQueueItem);
             this.callBumpEvents(backwardPositionInformation);
@@ -649,6 +651,19 @@ module.exports = class BaseEntity {
 
   reverseOffset(offset) {
     return [offset[0] * -1, offset[1] * -1];
+  }
+
+  handleMoveOffIronDoor(moveOffset) {
+    const formerPosition = [this.position[0] + moveOffset[0], this.position[1] + moveOffset[1]];
+    if (!this.controller.levelModel.inBounds(formerPosition[0], formerPosition[1])) {
+      return;
+    }
+
+    const wasOnDoor = this.controller.levelModel.actionPlane.getBlockAt(formerPosition).blockType === "doorIron";
+    const isOnDoor = this.controller.levelModel.actionPlane.getBlockAt(this.position).blockType === "doorIron";
+    if (wasOnDoor && !isOnDoor) {
+      this.controller.levelModel.actionPlane.findDoorToAnimate([-1, -1]);
+    }
   }
 
 };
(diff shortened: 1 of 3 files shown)


src/js/game/GameController.js
```diff
@@ -774,30 +760,15 @@ class GameController {
...
   moveDirection(commandQueueItem, direction) {
-    const entity = this.getEntity(commandQueueItem.target);
-    const moveOffset = this.directionToOffset(entity.movementState);
-
     this.execute(commandQueueItem, 'moveDirection', direction);
-
-    this.handleMoveOffIronDoor(commandQueueItem, moveOffset);
   }
 
   directionToOffset(direction) {
```
joshlory: Let's move this to `BaseEntity` now that it's only called there.

joshlory:
It looks like a second entity walking over the open door will cause it to close.  I don't think we need to fix this edge case right now, but adding a screen capture here to document the behavior:

door-close-conflict

src/js/game/Entities/BaseEntity.js
```diff
@@ -651,4 +653,17 @@ module.exports = class BaseEntity {
     return [offset[0] * -1, offset[1] * -1];
   }
 
+  handleMoveOffIronDoor(moveOffset) {
+    const formerPosition = [this.position[0] + moveOffset[0], this.position[1] + moveOffset[1]];
+    if (!this.controller.levelModel.inBounds(formerPosition[0], formerPosition[1])) {
```
joshlory: I'm just now realizing we have both `levelModel.inBounds(x, y)` and `levelPlane.inBounds(position)`.  Not something we need to clean up for this PR, but something we should address in the future!

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -355,6 +338,20 @@ module.exports = class LevelPlane {
     return posToRefresh;
   }
 
+  positionEquivalence(lhs, rhs) {
+    return (lhs[0] === rhs[0] && lhs[1] === rhs[1]);
+  }
+
+  checkEntityConflit(position) {
```
joshlory: Nit: `checkEntityConflict` misspelling.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -355,6 +338,20 @@ module.exports = class LevelPlane {
...
+    let captureReturn = false;
+    this.levelModel.controller.levelEntity.entityMap.forEach((workingEntity) => {
```
joshlory: This is a sign that `levelEntity` should probably live on `levelModel` instead of controller.  Not blocking this change.

joshlory: Please re-write with `.some()` or `.any()`

src/js/game/LevelMVC/LevelPlane.js
WilliamHarmonMS: 😞

WilliamHarmonMS:
@joshlory @Hamms I think this one is ready, if anyone has time. 👍

src/js/game/Entities/BaseEntity.js
```diff
@@ -651,4 +653,43 @@ module.exports = class BaseEntity {
...
+
+    const wasOnDoor = this.controller.levelModel.actionPlane.getBlockAt(formerPosition).blockType === "doorIron";
+    const isOnDoor = this.controller.levelModel.actionPlane.getBlockAt(this.position).blockType === "doorIron";
+    if (wasOnDoor && !isOnDoor) {
+      this.controller.levelModel.actionPlane.findDoorToAnimate([-1, -1]);
+    }
+  }
+
+
+
```
Hamms: nit: extra whitespace

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -355,6 +338,20 @@ module.exports = class LevelPlane {
     return posToRefresh;
   }
 
+  positionEquivalence(lhs, rhs) {
```
Hamms: nit: we've already got one of these defined in `GameController`. I vote we either use that or pull it out into a utility method

WilliamHarmonMS: @Hamms I wholeheartedly vote for a utility method. I want to say there are a number of odd helper functions throughout this codebase that could be pulled out into a utility class. I don't know if we care to do that now / later /whenever but I think that's a great scheme to work with.

joshlory:
Fixes https://github.com/code-dot-org/craft/issues/118.

Issue #164, opened 2017-10-02
Ambient occlusion shadows should be drawn above rails and redstone


joshlory:
Sample:

![image](https://user-images.githubusercontent.com/413693/31094947-c849d15c-a76b-11e7-96f8-bb4a9e2fdc0c.png)

joshlory:
Redstone example:

![image](https://user-images.githubusercontent.com/413693/31253757-e51a766c-a9da-11e7-992a-47a7420ac0a5.png)

![image](https://user-images.githubusercontent.com/413693/31253850-3052c1ca-a9db-11e7-8a1e-a2f9abb6089e.png)

joshlory:
I wonder if we should render the rails and redstone sprites as part of the `groundDecorationPlane` sprite group?

This would also solve the issue where redstone wire looks too flat, because it's not affected by https://github.com/code-dot-org/craft/blob/master/src/assets/images/Shade_Layer.png

![image](https://user-images.githubusercontent.com/413693/31412336-843c2762-adc9-11e7-9764-e81c5b10dadb.png)

joshlory:
This also affects Pressure Plates and Iron Doors:

<img width="400" src="https://user-images.githubusercontent.com/413693/31470197-935bd8f2-ae99-11e7-9b10-0d892350a42b.png">

Doors will require an angled shadow mask to be completely correct.

joshlory:
Mockup of what Pressure Plates should look like:

![image](https://user-images.githubusercontent.com/413693/31471414-7c4a1170-ae9d-11e7-8784-b6269d865066.png)

I created this screenshot by temporarily moving the Pressure Plates to the `groundDecorationPlane`, and the `groundDecorationPlane` sprites to render with the `groundPlane` group.

PR #275, opened 2017-10-12
Correct piston sort order for blocks in the level config

joshlory:
Partial fix for https://github.com/code-dot-org/craft/issues/164#issuecomment-335968741.  Currently we only correct piston sort order for newly-placed blocks.  Also correct piston sort order for blocks in the level config.

Before:

image

After:

image


joshlory:
This PR doesn't solve the (existing) problem that the Player's head is rendered below the piston arm:

image

Issue #85, opened 2017-09-07
No SFX occurs stepping on or off a Pressure Plate


v-samtha:
Found at: https://code-dot-org.github.io/craft/

Stepping onto a Pressure Plate does not cause the familiar Minecraft "click" sound effect to occur. No sound effect seems to occur instead.

There is also no sound effect when stepping off of the Pressure Plate as it does in Minecraft.

Steps:
1. (If necessary) Place Pressure Plate
2. Step on the Pressure Plate
3. Observe sound effects (if any)
4. Step off the Pressure Plate
5. Observe sound effcts (if any)

Attachment:
[HoC_PressurePlate_NoSFXSteppingOnOrOff.zip](https://github.com/code-dot-org/craft/files/1283051/HoC_PressurePlate_NoSFXSteppingOnOrOff.zip)

Issue #140, opened 2017-09-27
Pistons make no SFX when powered/unpowered


v-samtha:
Found at: https://code-dot-org.github.io/craft/

Giving power to a Piston causes the head of the piston to extend. However, when this animation occurs (be it when powering or unpowering), no sound effects play to alert the player that an action is occurring.

Steps:
1. Place a Piston
2. Hook up Redstone to the Piston
3. Power the Piston
4. Observe sound effects (if any)

Attachments:
[HoC_Piston_Activate_NoSFX.zip](https://github.com/code-dot-org/craft/files/1338750/HoC_Piston_Activate_NoSFX.zip)

PR #278, opened 2017-10-13
Play audio

WilliamHarmonMS:
@joshlory  or @Hamms PTAL

we're playing pressurePlate and Piston sound effects. I think that's it so far? Still working on music.

Fixes https://github.com/code-dot-org/craft/issues/85 and https://github.com/code-dot-org/craft/issues/140.


src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -330,14 +330,29 @@ module.exports = class LevelPlane {
...
+    switch (sound) {
+      case "On":
+        this.levelModel.controller.audioPlayer.play("pistonOut");
+        break;
+      case "Off":
+        this.levelModel.controller.audioPlayer.play("pistonIn");
+        break;
+      case "":
+        //do nothing
+        break;
```
joshlory: Tiny nit: you can remove this case.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -378,20 +393,24 @@ module.exports = class LevelPlane {
...
+    let returnValue = "";
 
     if (block.blockType.startsWith("piston") && !block.blockType.startsWith("pistonArm")) {
       block.isPowered = this.powerCheck(position, true);
       if (block.isPowered) {
         this.activatePiston(position);
+        returnValue = "On";
       } else if (!block.isPowered) {
         this.deactivatePiston(position);
+        returnValue = "Off";
```
joshlory: Can you change these "On" -> "Out" and "Off" -> "In" so you don't need the `playPistonSound` switch statement above?

joshlory:
Hmm, I can't seem to get the Piston to make a sound?

WilliamHarmonMS:
@joshlory 😞 

You're totally right, Pistons weren't being caught in some cases because I wasn't saving the string properly and it was getting set back to null before the sound call was made. That's fixed now. We should be good!

WilliamHarmonMS:
@joshlory PTAL fixed a merge conflict and this should be good to go ❤️

joshlory:
Hmm, placing a block next to the pressure plate seems to re-trigger the piston sound.

WilliamHarmonMS:
Good catch @joshlory 
I was trying to be clever and check for piston state changes, but I did it too high up and was playing the sound if pistons were powered / unpowered, regardless of what their -former- state was. This was in an effort to circumvent the playSound call being needed at the actual -activation / deactivation- calls, because if the sound gets called more than once (because more than 1 piston changes states as a result of the action) the sound got real bad real fast.

This is now fixed, but it either was going to involve passing return values up two or three functions, or just having a class-wide accessible flag that can be set in the activation/deactivation functions, but then only referenced and reset once after all the pistons are checked (this is the solution I went with).

WilliamHarmonMS:
Yeah they ended up being easily overlap-able, but this DOES prevent multiple version of the same sound from triggering at the same time (ie. if you power 4 pistons, only 1 piston sound is played, not 4x of the same audio)

PR #280, opened 2017-10-13
use isWalkable concept rather than skipHopAnimation

Hamms:


src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -248,6 +248,10 @@ module.exports = class LevelBlock {
     return this.isEmpty || this.isEntity;
   }
 
+  static isWalkable(blockType) {
+    return new LevelBlock(blockType).isWalkable;
+  }
```
Hamms: @joshlory I'm not sure how I feel about this approach to exposing block-level concepts; I'd love to hear what you think about it

joshlory: Not a huge fan... can we use `block.isWalkable` below?

Ideally we'd stop referencing `blockType` at all in the View, but that will take a bit of cleanup.

joshlory:
Cleanup looks great, but still on the fence about creating a new LevelBlock instance just to check some properties of a given `blockType`.  Seems like it would be much cleaner to always deal with `LevelBlock` instances in the View (but that will take a bit of cleanup).

src/js/game/LevelMVC/LevelBlock.js
Hamms: Agreed that ideally we'd do that, but as it current stands there is nowhere in the call stack of this method that is actually creating a LevelBlock for us to call `isWalkable` on

PR #281, opened 2017-10-13
Add some visual variance to the Action Plane, to match Shade_Layer.png

joshlory:
The Ground Plane uses ["Shade_Layer.png"](https://github.com/code-dot-org/craft/blob/b7a639891326a1a054280e865de5501353cda91a/src/assets/images/Shade_Layer.png) to provide some visual variation and make the grid square visible.  Add similar variation to the Action Plane.

Before:

image

After:

image


joshlory:
For comparison, here's what the Ground Plane would look like with no variation:

image

And here's what it actually looks like: (using [Shade_Layer.png](https://github.com/code-dot-org/craft/blob/b7a639891326a1a054280e865de5501353cda91a/src/assets/images/Shade_Layer.png), not random sprite tint)

image

PR #282, opened 2017-10-13
Replace a few blockType comparisons with block methods

Hamms:


joshlory:
Looks like npm-debug.log got checked in accidentally.

joshlory:
I like this approach!

Out of curiosity, what are the pros/cons for using instance methods like `getIsPiston()` vs. setting `this.isPiston = true` in the constructor?

WilliamHarmonMS:
In regards to getIsPiston() vs this.isPiston I would say that currently, isPiston is read/WRITE so the value can change, where as an embedded methods like getIsPiston() has a reliable value that can't be changed? If we had a const isPiston = true; in the constructor that might be different, but even then perhaps readability? If someone see's levelPlane referencing this.isPiston they don't -know- it's read only just from looking at it, where as they can confidently assume that getIsPiston() will always return a consistent value?

That'd be my guess?

Hamms:
I'd say there are also two important code quality considerations

First, readability. Right now, to determine which kind of blocks are, for example, `walkable`, you have to search through the constructor method for all if statements that are changing `isWalkable` from the default. And for things like `isConnectedToRedstone`, you [sometimes have to go even deeper](https://github.com/code-dot-org/craft/blob/master/src/js/game/LevelMVC/LevelBlock.js#L63). Whereas concepts implemented using `get` methods are always readable [even when relatively complex](https://github.com/code-dot-org/craft/blob/master/src/js/game/LevelMVC/LevelBlock.js#L239-L245).

Second, extendability. We want to eventually implement a hierarchy of LevelBlock classes, and overriding methods is much nicer than overriding constructor logic in that pattern.

Issue #285, opened 2017-10-13
The [D] key no longer moves Steve in Playground


v-samtha:
Found at: https://code-dot-org.github.io/craft/

In the Playground, a player can use WASD to move Steve around. However, the [D] key does not move Steve to the right.

Note, this does not seem to be an issue in the actual levels (https://levelbuilder-studio.code.org/s/colehoc2017/stage/1/puzzle/1). The [D] key works there.

Steps:
1. Load the Playground
2. Press the D key
3. Observe lack of Steve movement

Attachments:
![ezgif-4-cc6383da20](https://user-images.githubusercontent.com/31709758/31558076-553e447a-b000-11e7-8339-a4d0771e2b1e.gif)

PR #286, opened 2017-10-13
ascii values don't go up to 868... woops

WilliamHarmonMS:
Fixes https://github.com/code-dot-org/craft/issues/285.


WilliamHarmonMS:
@joshlory 
68 == 'D' which is what we want?
100 == 'd' but since the others seem to be referencing the capital letters, that's what I went with.

Issue #288, opened 2017-10-13
Sticky Piston pushing causes wraparound in Level 5


v-samtha:
Found at: https://levelbuilder-studio.code.org/s/colehoc2017/stage/1/puzzle/5

In Level 5, the Agent maneuvers through the track, stepping on Pressure Plates to activate Sticky Pistons to push and pop blocks to allow a pathway for Steve to reach the victory block. However, the first Pressure Plate actives a Sticky Piston that pushes a Grass block of the map. This actually causes it to wrap around to the row above it, breaking the Redstone Wire line that would allow Steve to get the Diamond Block.

Steps:
1. Load Level 5
2. Code the Agent to move to the first Pressure Plate
3. Observe pushed Grass Block movement

Attachments:
![ezgif-4-3e3ed45dc5](https://user-images.githubusercontent.com/31709758/31561806-ef71387e-b00d-11e7-94c5-574f503ed17a.gif)

PR #292, opened 2017-10-13
woooops. We weren't checking bounds for off-map-block-pushing

WilliamHarmonMS:
@joshlory or @Hamms 

Goofed and didn't check bounds. setBlockAt() calls coordinatesToIndex() which assumes everything is in bounds, and it ends up wrapping: [-1, 3] is interpreted as [9, 2] (because y * width + x, if x is -1 you just end up with 30 - 1 = 29)

Completely happenstance that it wrapped perfectly to the row above. This was breaking the diamond path in level 5, but should be fine now.

Resolves https://github.com/code-dot-org/craft/issues/288.

### DIFF
--- a/src/js/game/LevelMVC/LevelPlane.js
+++ b/src/js/game/LevelMVC/LevelPlane.js
@@ -612,7 +612,9 @@ module.exports = class LevelPlane {
         }
         redo = true;
       }
-      this.setBlockAt(destination, this.getBlockAt(blocksPositions[i]));
+      if (this.inBounds(destination)) {
+        this.setBlockAt(destination, this.getBlockAt(blocksPositions[i]));
+      }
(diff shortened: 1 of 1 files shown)


src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -612,7 +612,9 @@ module.exports = class LevelPlane {
         }
         redo = true;
       }
-      this.setBlockAt(destination, this.getBlockAt(blocksPositions[i]));
+      if (this.inBounds(destination)) {
+        this.setBlockAt(destination, this.getBlockAt(blocksPositions[i]));
```
joshlory: Hmm, or should `setBlockAt` verify that the destination is in bounds?

PR #293, opened 2017-10-13
Replace shade layer with tint

joshlory:
Follow up to PR https://github.com/code-dot-org/craft/pull/281.

Ground plane before: (using Shade_Layer.png)

image

Ground plane after: (using random tint)

screen shot 2017-10-13 at 1 40 08 pm


src/js/game/LevelMVC/LevelView.js
```diff
@@ -1971,6 +1978,7 @@ module.exports = class LevelView {
         sprite = group.create(xOffset + 40 * x, yOffset + group.yOffset + 40 * y, atlas, frame);
         frameList = Phaser.Animation.generateFrameNames("Lava_", 0, 5, "", 0);
         sprite.animations.add("idle", frameList, 5, true);
+        this.psuedoRandomTint(group, sprite, x, y);
```
Hamms: why are these handled specially?

joshlory: Because this switch method has lots of duplicated code? 😄

It would be worth refactoring `createBlock` at some point to clean it up.

Hamms: Hah, fair. But still, why are these specifically handled specially, when it doesn't look like the old approach was treating them special?

joshlory: Ah, sorry, I see what you're asking!  The old approach indiscriminately place `Shade_Layer.png` over the entire ground group, regardless of block type.  To duplicate that effect we have to apply the tint to all ground block types.

Hamms: ahh, and these two are the ones that can end up as just regular ground blocks but which aren't caught by the `default`, gotcha.

In that case, we probably want to also do this for "watering"

joshlory: Followed up offline.  This actually calls back into the same function (`createBlock`) with "farmlandWet", which falls into the default case below.

https://github.com/code-dot-org/craft/pull/293/files#diff-894dfbadc994764b5b510d470f495021L1960

PR #295, opened 2017-10-13
Fix ground group refresh bug

Hamms:
Because `refreshGroundGroup` is now partially responsible for populating LevelView's `actionPlaneBlocks` array, when refreshing the action group we don't want to reset `actionPlaneBlocks` until after refreshing the ground group. And because blocks are no longer added to the array strictly in order (because we add all ground blocks first in the `refreshGroundGroup` method before going in and adding the action blocks), it no longer makes sense to use `.push` to update the array.

And finally, since we're no longer using `.push`, the creation of the blocks that happens in `refreshGroups` is now practically identical to the creation of blocks that happens in the `createActionPlaneBlock` helper method, so just use that instead of any custom logic.

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -1012,17 +1012,32 @@ module.exports = class LevelView {
   }
 
   createActionPlaneBlock(position, blockType) {
-    let blockIndex = (this.yToIndex(position[1])) + position[0];
+    const block = new LevelBlock(blockType);
+    const blockIndex = (this.yToIndex(position[1])) + position[0];
+    if (block.isEmpty) {
+      this.actionPlaneBlocks[blockIndex] = null;
+      return;
+    }
 
     // Remove the old sprite at this position, if there is one.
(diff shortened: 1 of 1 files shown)


src/js/game/LevelMVC/LevelView.js
```diff
@@ -1012,17 +1012,33 @@ module.exports = class LevelView {
   }
 
   createActionPlaneBlock(position, blockType) {
-    let blockIndex = (this.yToIndex(position[1])) + position[0];
+    const block = new LevelBlock(blockType);
+    const blockIndex = (this.yToIndex(position[1])) + position[0];
 
     // Remove the old sprite at this position, if there is one.
     this.actionGroup.remove(this.actionPlaneBlocks[blockIndex]);
     this.groundGroup.remove(this.actionPlaneBlocks[blockIndex]);
```
joshlory: What makes this not remove the block beneath rails by accident?  It's a no-op when the target sprite isn't a member of the `groundGroup`?

Hamms: precisely! `group`s are set-style containers

PR #304, opened 2017-10-16
Ghast entity

WilliamHarmonMS:
The Ghast is in and can sit there and animate idle. NO SHOOTING YET


WilliamHarmonMS:
@joshlory  or @Hamms 

PTAL... (This may be as far as we go with it. I think we've axed the "shooting" mechanic, but there are still parties quite interested in bolstering the theming of The Nether with this additional entity)

joshlory:
Should the Ghast be floating above the action plane?  Right now collision is a bit strange:

ghast-collision

joshlory:
No, I think this is fine for now.

Issue #290, opened 2017-10-13
Bedrock blocks don't cast shadows


joshlory:
See: https://code-dot-org.github.io/craft/?level=agent12

![image](https://user-images.githubusercontent.com/413693/31563965-055096fa-b016-11e7-818e-b523b83a8845.png)

joshlory:
Expected:

![image](https://user-images.githubusercontent.com/413693/31631385-69d9be76-b26e-11e7-9c36-d8c2e64c0c6c.png)

PR #309, opened 2017-10-16
Bedrock should cast shadows

joshlory:
Fixes https://github.com/code-dot-org/craft/issues/290.

It looks like "bedrock" was mistakenly classified as an entity block in PR https://github.com/code-dot-org/craft/pull/153.

### DIFF
--- a/src/js/game/LevelMVC/LevelBlock.js
+++ b/src/js/game/LevelMVC/LevelBlock.js
@@ -172,7 +172,7 @@ module.exports = class LevelBlock {
       this.isEntity = true;
     }
 
-    if (blockType.startsWith("bed")) {
+    if (blockType === "bedFoot" || blockType === "bedHead") {
       this.isEntity = true;
     }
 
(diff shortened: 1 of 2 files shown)


src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -172,7 +172,7 @@ module.exports = class LevelBlock {
       this.isEntity = true;
     }
 
-    if (blockType.startsWith("bed")) {
+    if (blockType === "bedFoot" || blockType === "bedHead") {
```
Hamms: As as aside, this is exactly why I hate that we're relying so much on string comparisons for all this

joshlory: Agreed!

WilliamHarmonMS: Does this change anything? Are there any other blocks other than Foot and Head that begin with 'bed' string?

joshlory:
Worse than nothing — it caught the "bedFoot" and "bedHead" cases, but also "bedrock" by mistake.

src/js/game/LevelMVC/LevelBlock.js
joshlory: Yes: "bedrock".

Issue #209, opened 2017-10-04
Victory animation (and level completion) interrupted on Redstone change


v-samtha:
Found at: https://levelbuilder-studio.code.org/s/colehoc2017/stage/1/puzzle/2

A player successfully navigates Steve to a victory block. During the victory animation (where the chime sfx plays and and they jump up and down), a Redstone change occurs. For example, in Level 2, the Agent standing on the Pressure Plate moves off of it while the animation is playing. This not only interrupts the victory animation, but the game does not progress to the level completion screen and causes Steve to become stuck (until reset)

Steps:
1. Load Level 2
2. Program the Agent to either Pressure Plate
3. Have the Agent idle (spin in place)
4. Navigate Steve to the victory block
5. During the animation, have the Agent move off the Pressure Plate (may take some coding tweaks)
6. Observe victory animation and level completion

Attachments:
[HoC_VictoryAnimation_InterruptedOnRedstoneChange.zip](https://github.com/code-dot-org/craft/files/1357552/HoC_VictoryAnimation_InterruptedOnRedstoneChange.zip)

joshlory:
Tracking all animation interruption bugs under this issue.

Issue #267, opened 2017-10-11
When both Steve and the Agent strike/destroy at the same time player losses control of Steve.


v-jaarba:
Found at: https://levelbuilder-studio.code.org/s/colehoc2017/stage/1/puzzle/11

Whenever Steve and the Agent are at the same location:
The Agent bumps into the block while Steve attempts attack/use on a block the player loses all control of Steve.

**Repro Steps:**
1) Start level 11.
2) Create a loop that moves the agent and destroys blocks until it hits a block  (Keep the loop going so Steve can get to the same location).
3) Start the simulation.
4) Move Steve to the same location the agent is in.
5) Hit the space bar to use/attack the block.

**Expected Results:**
The player does not lose control of Steve.

**Actual Results:**
The Player does lose control of Steve.

**Notes:**
The wall around the map also reproes this issue.

**Video:**
Shows this happen on the wall but it also happens on any indestructible block on the map.
[Microsoft Edge 10_11_2017 2_37_54 PM.zip](https://github.com/code-dot-org/craft/files/1377281/Microsoft.Edge.10_11_2017.2_37_54.PM.zip)

joshlory:
Tracking under issue https://github.com/code-dot-org/craft/issues/209.

Issue #261, opened 2017-10-11
Player cannot finish level until agent completes algorithm of moving


LB-wokuja:
Found at: https://levelbuilder-studio.code.org/s/colehoc2017/stage/1/puzzle/3
Found also at puzzles from 4 to 10.
 
Repro Steps:
Create algorithm, that allows you to pass level, and then add a looped action of moving ( That will not collide with possibility of finishing level).
Move Steve/Alex to the torch block, before agent will finish algorithm.
Observe issue.

Attached file:
[Player cannot end level.zip](https://github.com/code-dot-org/craft/files/1376021/Player.cannot.end.level.zip)

joshlory:
Good catch!

The Player reaching the finish square should immediately end the level, even if the Agent still has commands left in the queue.  @WilliamHarmonMS it also looks like the Agent bump animation is interrupting the Player animation.

joshlory:
Possible duplicate of issue https://github.com/code-dot-org/craft/issues/209.

WilliamHarmonMS:
@joshlory it 100% looks like a bump animation is what's doing it, and I have a fix for that in PR#268.

As for redstone charge propagation.... that's a bit trickier. That shouldn't be effecting the player's animation at all? Unless something is getting tripped up in the plane refreshing? Which would be odd.

Can we confirm that the charge....

Never mind. It's the door thing. Iron doors set player animation back to idle. That's got to be it.

joshlory:
I'm going to close this as a duplicate of https://github.com/code-dot-org/craft/issues/209.  Feel free to re-open if it turns out to be a different underlying issue.

PR #310, opened 2017-10-16
iron doors no longer set player animation to idle

WilliamHarmonMS:
fixes issue: [#209](https://github.com/code-dot-org/craft/issues/209)


WilliamHarmonMS:
@joshlory  or @Hamms PTAL

joshlory:
Does this also fix both the linked issues (https://github.com/code-dot-org/craft/issues/261 and https://github.com/code-dot-org/craft/issues/267)?  Or should I re-open those to track?  Thanks!

WilliamHarmonMS:
@joshlory #261 was very general and actual encompassed two instances where Agent actions were setting player animation back to idle. I fixed the 'bump' instance in another PR, but this is another case where the iron door was doing more or less then same thing. For destroy though (#267 there) that might actually be ANOTHER instance of this same phenomenon that I haven't witnessed. I'll go hunt that down and see if I can find anything.

PR #322, opened 2017-10-17
Let the player fall into water & lava in Agent levels

joshlory:
We have these great animations for falling in water & lava!  Let's use them.


Hamms:
My understanding is that we only want to fall in to the water and lava when the goal of the level is to program the player, and encountering water

PR #323, opened 2017-10-17
Save correct miniblock offset

### LINKED ISSUE #319: Player can collect diamond by standing on the block one square above the diamond ore block
Found at: [https://levelbuilder-studio.code.org/s/colehoc2017/stage/1/puzzle/8](https://levelbuilder-studio.code.org/s/colehoc2017/stage/1/puzzle/8)

It is not necessary for the player to step on the diamond ore square to collect the diamond.  By stepping on the square above the diamond ore, the player collects the diamond.  For several levels this cannot be done because there is a block preventing the player from standing in the square above the diamond ore block.

1. Load Level 5, 8, 10, or 11
2. Program the agent to clear or build a path to the diamond
3. Move the player to the square above the diamond ore block
4. Observe

Actual result: Player picks up diamond without stepping on diamond ore block
Expected result: Player picks up diamond only by stepping on diamond ore block (or one square away in any direction up, down, left, or right)

[level8diamond.zip](https://github.com/code-dot-org/craft/files/1392033/level8diamond.zip)

Hamms:
Collectible minblocks were rendering with the actionGroup yOffset being
taken into account, but were not including that offset when saving
themselves to collectibleItems, meaning that the distance calculations
ultimately being used to determine whether or not the player is close
enough to pick them up were off by just that much.

Fixes #319

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -1858,10 +1858,10 @@ module.exports = class LevelView {
     const frameStart = this.miniBlocks[frame][1];
     const frameEnd = this.miniBlocks[frame][2];
     const xOffset = -10 - (xOffsetRange / 2) + (Math.random() * xOffsetRange);
-    const yOffset = 0 - (yOffsetRange / 2) + (Math.random() * yOffsetRange);
+    const yOffset = 0 - (yOffsetRange / 2) + (Math.random() * yOffsetRange) + this.actionGroup.yOffset;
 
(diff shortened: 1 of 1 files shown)


joshlory:
Worth adding a LevelView unit test or integration test?

Hamms:
I attempted to add one, but unfortunately a lot of this logic happens in the animation onComplete callback for the little bounce that the miniblock does, and the tests as we currently have them don't fully run the animations. I don't think it's worth it to add all that functionality just to test this

Issue #313, opened 2017-10-16
Parity/Unitended results - Redstone torches placed in water will be destroyed in Minecraft.


v-jaarba:
Found in:https://levelbuilder-studio.code.org/s/colehoc2017/stage/1/puzzle/14

In Minecraft if a player ends up placing a torch in water it will be destroyed instantly.
In HoC the placing of a Redstone torch will cause the water to be pushed away and the player can then walk across the gap created in the water, effectively walking on top of the Redstone torch. 

**Repro Steps:**
1) Start HoC on mission 14.
2) Setup the Agent to move on top of the water and then place a Redstone Torch on to the water.
3) Observe the water is moved out of the index.

**Expected Results:**
The water would destroy the Redstone torch or the player could not place the torch in water.

**Actual Results:**
The water is no long in the index allowing the player to walk on top of the Redstone torch.

**Note:**
This is not Parity of Minecraft and it causes an unintended side effect of being able to cross rivers and gaps with an item Steve/Alex should not be able to walk on.

**Video:**
[Walking on Redstone Torch.zip](https://github.com/code-dot-org/craft/files/1389330/Walking.on.Redstone.Torch.zip)

PR #327, opened 2017-10-17
No torches in water

Hamms:
Fix https://github.com/code-dot-org/craft/issues/313

### DIFF
--- a/src/js/game/LevelMVC/LevelBlock.js
+++ b/src/js/game/LevelMVC/LevelBlock.js
@@ -232,6 +232,24 @@ module.exports = class LevelBlock {
   }
 
   /**
+   * Can this block be placed in liquid to replace a liquid block? Should
+   * generally be true for all "standard" blocks like cobblestone and dirt, and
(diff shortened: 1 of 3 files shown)


src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -232,6 +232,24 @@ module.exports = class LevelBlock {
   }
 
   /**
+   * Can this block be placed in liquid to replace a liquid block? Should
+   * generally be true for all "standard" blocks like cobblestone and dirt, and
+   * false for all "special" blocks like redstone and torches.
+   *
+   * @return {boolean}
+   */
+  getIsPlaceableInLiquid() {
```
joshlory: I wonder if we want to generalize the idea of a normal "block" vs. non-blocks.  Whether not something is a block has implications about whether or not it's walkable, etc.  I _think_ all walkable blocks can't be placed in water?

But this is great first-pass cleanup and bug-fix!

PR #329, opened 2017-10-17
new agent spawn location at the end of level 1

WilliamHarmonMS:
@joshlory @Hamms 

Due to feedback, we're looking to change the Agent spawn location at the end of level 1 from above the player to below the player.


joshlory:
I wonder if the Agent should spawn to the east of the player, instead of to the south.  Can you share a screenshot of the updated level?

WilliamHarmonMS:
The new Agent spawn location was specifically requested to be in below the player (as if it were leading the way out the door) by Deirdre and Cole yesterday afternoon. I'll fix this darn Agent test though 😭  

didn't foresee that one.

PR #331, opened 2017-10-18
Redstone sparkle animation

joshlory:
Only includes the `groundDecorationPlane` animation.  No logic to dynamically add and remove the sprite when redstone is powered and unpowered.

redstone-sparkle


joshlory:
Merging this for starters, we can add follow-up PRs to address any design feedback!

Issue #244, opened 2017-10-09
Miniblocks not picked up when rolling over them in Minecart


v-samtha:
Found at: https://levelbuilder-studio.code.org/s/colehoc2017/stage/1/puzzle/10

In Level 10, the player uses Steve to board a Minecart and ride through the tracks to victory. However, if Steve passes near or over miniblocks, Steve does not automatically pick them up as he does in Minecraft.

Steps:
1. Load Level 10
2. Write code to solve the puzzle via Minecart
3. Have Steve use the Minecart
4. Observe miniblocks on the ground

Attachments:
![ezgif-1-9c4cfea809](https://user-images.githubusercontent.com/31709758/31360809-f674d010-ad04-11e7-8d6e-2d83c48ca6b0.gif)

joshlory:
Nice find!  It also looks like the z-index of the minecart is a bit off (renders above the wall to the south in the beginning of the gif).

@WilliamHarmonMS is this a side effect of not changing the indicator and player location while animating the minecart?

Issue #238, opened 2017-10-06
No way to exit a minecart when in a rail loop


joshlory:
Can't leave the minecart, have to restart the level.

![minecart-loop](https://user-images.githubusercontent.com/413693/31296263-84779eae-aa96-11e7-9aca-789e24ee7491.gif)

WilliamHarmonMS:
fixed in [PR#332](https://github.com/code-dot-org/craft/pull/332)

PR #332, opened 2017-10-18
Hopping on rails + Handle corner animation

WilliamHarmonMS:
@joshlory or @Hamms 

This PR allows players to pickup miniblocks while on the rails AND allows the Rails to be ridden simply by walking onto them (a change I put in only because with the 'use' taken out, I couldn't hop on rails to test the pick up functionality!)

Fixes https://github.com/code-dot-org/craft/issues/244 and https://github.com/code-dot-org/craft/issues/238.


WilliamHarmonMS:
addresses [issue#244](https://github.com/code-dot-org/craft/issues/244)

joshlory:
I'm having trouble crossing tracks at right angles.  If I hold down an arrow key and walk towards the track, the Player pauses in front of the track.

The Player can also start riding the rails from a corner — is that intended?

WilliamHarmonMS:
The corner rails I think is fine, but will confirm with design.

as for the perpendicular rails stopping movement, that's an excellent catch. It must be something with the moveDirection seeing there is a rail there, so not pushing the move command like normal, but also not being able to handle the rail. So.. Gotta look at that for sure. Will finish that first thing tomorrow.

WilliamHarmonMS:
@joshlory 

So I changed the way I check for rails to hop on. This prevents you from stopping when there is an invalid rail connection (so if you approach perpendicularly you'll just keep walking). It basically evaluates your move direction and asks if the rail you're going to move into as a further connection in that same direction. If so: ride, if not: walk like normal.

This feel good to you?

joshlory:
Can we separate this out into two PRs?  The item pickup change looks great, and I don't want to block it on discussion of the movement changes.

WilliamHarmonMS:
@joshlory  or @Hamms 
PTAL

I have a feeling you guys won't love this structure, but between crashes and borked functionality, this is where I ended up for covering all cases correctly.

At the end of the day, we DO have desired functionality, even if it's not pretty 😭

WilliamHarmonMS:
Changed the name of this PR because that's now more the focus. The item pick up is in the other PR.

WilliamHarmonMS:
@Hamms or @joshlory 

(this one specifically is dependent on PR#345)

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -261,6 +261,46 @@ module.exports = class LevelModel {
...
+  }
+
+  isNextRailValid(block, direction) {
+    if (!block) {
+      return;
+    }
+    return this.inverseDirection(block.connectionA) === direction ||
+    this.inverseDirection(block.connectionB) === direction ||
+    block.connectionA === direction ||
+    block.connectionB === direction;
```
Hamms: nit: indent hanging lines

```javascript
return this.inverseDirection(block.connectionA) === direction ||
  this.inverseDirection(block.connectionB) === direction ||
  block.connectionA === direction ||
  block.connectionB === direction;
```

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -261,6 +261,46 @@ module.exports = class LevelModel {
...
+  }
+
+  inverseDirection(direction) {
+    return (direction + 2) % 4;
+  }
```
Hamms: I believe there is already a FacingDirection method for this

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -261,6 +261,46 @@ module.exports = class LevelModel {
...
+        break;
+      case 2:
+        offset[1] = 1;
+        break;
+      case 3:
+        offset[0] = -1;
+        break;
+    }
+    return [entity.position[0] + offset[0], entity.position[1] + offset[1]];
+  }
```
Hamms: This can all simply be `Position.add(entity.position, FacingDirection.directionToOffset(entity.movementState))`

src/js/game/LevelMVC/LevelView.js
```diff
@@ -769,6 +782,18 @@ module.exports = class LevelView {
     }
   }
 
+  /**
+  * Handling the first case of walking onto a track while not currently on one
+  */
+  firstTime(currPos, nextPos) {
```
Hamms: nit: this method should be much more specifically-named; something like `isFirstTimeOnRails`

src/js/game/LevelMVC/LevelView.js
```diff
@@ -769,6 +782,18 @@ module.exports = class LevelView {
...
+    let nextBlock = this.controller.levelModel.actionPlane.getBlockAt(nextPos);
+    let currBlock = this.controller.levelModel.actionPlane.getBlockAt(currPos);
+    if (!currBlock.blockType.startsWith("rails") && nextBlock.blockType.startsWith("rails")) {
```
Hamms: these should both _definitely_ be using `isRails` to avoid `railsRedstoneTorch`

src/js/game/Entities/Player.js
```diff
@@ -88,6 +95,9 @@ module.exports = class Player extends BaseEntity {
     const queueHasOne = this.queue.currentCommand && this.queue.getLength() === 0;
     const timeEllapsed = (+new Date() - this.lastMovement);
     const movementAlmostFinished = timeEllapsed > 300;
+    if (isMoving && timeEllapsed > 800) {
```
Hamms: what's the reasoning for this time limit? I'd love to see a comment explaining

src/js/game/Entities/Player.js
```diff
@@ -10,6 +10,9 @@ module.exports = class Player extends BaseEntity {
     this.inventory = {};
     this.movementState = -1;
     this.onTracks = false;
+    this.getOffTrack = false;
+    this.firstTrack = true;
+    this.railDelayCheck = false;
```
Hamms: `firstTrack` and `railsDelayCheck` appear to be unused

WilliamHarmonMS: Good call. artifacts from when I was trying to get the right behavior. 👍

src/js/game/Entities/Player.js
WilliamHarmonMS: Fair, I'm not a fan of magic numbers either, but I was essentially using the same logic as the movementAlmostFinished check. Waiting for one command to finish before checking for others so that functionality doesn't get tangled.

src/js/game/LevelMVC/LevelModel.js
WilliamHarmonMS: So it does. But ppfffffft, why switch case when you can just add and mod? 😂

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: isRails doesn't seem to be a thing currently? I can make a helper for that though 👍

src/js/game/LevelMVC/LevelModel.js
WilliamHarmonMS: that as a singular statement won't always work because once the player is on the rails (past the first index), they're being moved due to the rail tracking logic, not a moveDirection call, which means it's possible for their movementState to be -1 which will return undefined from the directionToOffset call, which will cause a crash further down the line.

Using the switch statement above and calling:
return Position.add(entity.position, offset);

will mean that if movementState was -1 we'll simply fall through and offset will be [0,0] and everything works as expected. So this is what I've done 🙆‍♂️

WilliamHarmonMS:
@Hamms another iteration! It looks like we're all green and still functioning appropriate so I -believe- it's been rebased on top of the master branch correctly, and that conflicts have been resolved.

src/js/game/LevelMVC/FacingDirection.js
```diff
@@ -26,11 +26,11 @@ const FacingDirection = Object.freeze({
     return (facing + 4 + (rotation === 'right' ? 1 : -1)) % 4;
   },
 
-  directionToOffset(direction) {
+  directionToOffset: function (direction) {
     switch (direction) {
       case FacingDirection.North: return [0, -1];
-      case FacingDirection.South: return [0, 1];
       case FacingDirection.East: return [1, 0];
+      case FacingDirection.South: return [0, 1];
```
Hamms: super-minuscule nit: it'd be nice to have the order of the directions match all the other `switch`es in this file

src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -209,6 +209,10 @@ module.exports = class LevelBlock {
         this.blockType.startsWith("pressurePlate");
   }
 
+  isRails() {
+    return this.blockType.startsWith("rails") && this.blockType !== "railsRedstoneTorch";
+  }
+
```
Hamms: oh, sorry for my typo; I meant use the existing `isRail` flag

src/js/game/LevelMVC/LevelView.js
```diff
@@ -714,16 +714,29 @@ module.exports = class LevelView {
...
   playTrack(position, facing, isOnBlock, entity = this.player, completionHandler) {
     entity.onTracks = true;
-    const track = this.controller.levelModel.actionPlane.getMinecartTrack(position, facing);
 
-    if (!track) {
+    // Need to get track on current position to avoid mishandling immediate turns
+    let track = this.controller.levelModel.actionPlane.getMinecartTrack(position, facing);
+
+    let offset = FacingDirection.directionToOffset(facing);
+    let nextPos = [entity.position[0] + offset[0], entity.position[1] + offset[1]];
```
Hamms: this can use `Position.add`, or if we're not using `offset` elsewhere it can just be `let nextPos = Position.forward(entity.position, facing)`

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -258,6 +258,42 @@ module.exports = class LevelModel {
...
+        break;
+      case 2:
+        offset[1] = 1;
+        break;
+      case 3:
+        offset[0] = -1;
+        break;
+    }
+    return Position.add(entity.position, offset);
+  }
```
Hamms: (replying to https://github.com/code-dot-org/craft/pull/332#discussion_r147794109 )

I'd still rather avoid duplicating so much code. How about just:
```javascript
const offset = FacingDirection.directionToOffset(entity.movementState) || [0, 0];
```

src/js/game/LevelMVC/LevelBlock.js
WilliamHarmonMS: d'oh, I totally didn't even see that and assumed we wanted a method for it. 😂 

my bad.

src/js/game/LevelMVC/LevelModel.js
WilliamHarmonMS: That works too. 🙆‍♂️

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: @Hamms Gotta love these new Position methods ❤️

WilliamHarmonMS:
@Hamms Alright! Next iteration is up, take a look and let me know what you think.

PR #336, opened 2017-10-19
Corner shadows

joshlory:
I modified the unused "Shadow_Parts_Fade_overlap" frame of the "Block_Shadows.png" asset.  Draw angled shadows to blocks with a non-walkable southwest neighbor.  Still need to run this change by Joel...

Before:

corner-shadows-before

After:

corner-shadows-after


joshlory:
Before:

corner-shadows-before-2

After:

corner-shadows-after-2

joshlory:
This is ready to merge, if Joel is ok with the style changes.

I _think_ this is probably close to the intended look from 2015.  I say that for two reasons:
 - There's a slightly different version of the asset (the one I modified) called "Shadow_Parts_Fade_overlap".  It looks like it may have been intended as the corner shadow.
 - Most places that would've exhibited this issue in 2015 have a `tallGrass` placed over the missing corner shadow.  I wonder if they didn't have time to fully-implement the corner shadows, and added the grass to cover it up where possible.

image

image

Issue #318, opened 2017-10-17
Agent is able to place redstone wire floating on top of presure plates.


v-jaarba:
Found In: https://levelbuilder-studio.code.org/s/colehoc2017/stage/1/puzzle/14

This may lead to future bugs down the road. In Minecraft Redstone wire is not able to be placed on top of pressure plates. But in HoC the Agent can place Redstone wire floating on top of pressure plates.

**Repro Steps:**
1) Start level 14.
2) Set up the agent to walk over a water tile then place a pressure plate down on the water. (works just like Minecraft)
3) Then have the agent place Redstone wire down over the pressure plate and have him move off the tile.
4) Observe.

**Expected Results:**
The Redstone wire is not able to be placed on top of the pressure plate which follows parity and won't look odd.

**Actual Results:**
The Redstone wire can be placed on top of pressure plates which then looks odd as it seems to be floating.

**Notes:**
Going forward if we ever can use Redstone torches/wire the Redstone mechanics need to closely follow the same mechanics Minecraft uses. Plus, it looks odd to have Redstone wires floating a block level above the pressure plates.

**Video:**
[RedstonewireonPressure.zip](https://github.com/code-dot-org/craft/files/1392025/RedstonewireonPressure.zip)

PR #338, opened 2017-10-19
can not put pressure plates on land but not in water/lava

WilliamHarmonMS:
@joshlory PTAL

This fixes [issue:318](https://github.com/code-dot-org/craft/issues/318)

pressurePlates should be fine to place on land (and therefore on the action Plane) but NOT over water, thus in the ground plane, thus allowing the player to put redstoneWire over top of them. 

What a silly mistake 😥

### DIFF
--- a/src/js/game/Entities/Agent.js
+++ b/src/js/game/Entities/Agent.js
@@ -25,9 +25,7 @@ module.exports = class Agent extends BaseEntity {
         result.plane = "groundPlane";
       }
     } else {
-      if (toPlaceBlock.isRedstone ||
-      toPlaceBlock.isRail ||
-      toPlaceBlock.blockType === "pressurePlateUp") {
+      if (toPlaceBlock.isFlat() ) {
         result.canPlace = true;
         result.plane = "actionPlane";
       }
--- a/src/js/game/Entities/Agent.js
+++ b/src/js/game/Entities/Agent.js
@@ -25,7 +25,9 @@ module.exports = class Agent extends BaseEntity {
         result.plane = "groundPlane";
       }
     } else {
-      if (toPlaceBlock.isRedstone || toPlaceBlock.isRail) {
+      if (toPlaceBlock.isRedstone ||
+      toPlaceBlock.isRail ||
+      toPlaceBlock.blockType === "pressurePlateUp") {
         result.canPlace = true;
         result.plane = "actionPlane";
       }
--- a/src/js/game/Entities/Agent.js
+++ b/src/js/game/Entities/Agent.js
@@ -25,7 +25,7 @@ module.exports = class Agent extends BaseEntity {
         result.plane = "groundPlane";
       }
     } else {
-      if (toPlaceBlock.isFlat() ) {
+      if (toPlaceBlock.isWalkable) {
         result.canPlace = true;
         result.plane = "actionPlane";
       }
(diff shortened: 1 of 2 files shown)


src/js/game/Entities/Agent.js
```diff
@@ -25,7 +25,9 @@ module.exports = class Agent extends BaseEntity {
         result.plane = "groundPlane";
       }
     } else {
-      if (toPlaceBlock.isRedstone || toPlaceBlock.isRail) {
+      if (toPlaceBlock.isRedstone ||
+      toPlaceBlock.isRail ||
+      toPlaceBlock.blockType === "pressurePlateUp") {
```
joshlory: Why do we also need this change?  Shouldn't the `getIsPlaceableInLiquid` fix be enough?

WilliamHarmonMS: getIsPlaceableInLiquid() change is sufficient to eliminate the ability to place the pressurePlate in the ground plane, which drops into that else statement, but the subsequent if, since it's not redstone or a rail, fails and without the additional inclusion of the pressurePlateUp blocktype check the Agent won't place anything at all on the actionPlane when over ground.

I am like 99% that the spec is that the Agent should be able to place pressurePlates, just like redstone torches, over ground as a source of power for redstoneWire.

WilliamHarmonMS: A better change may be, perhaps, a new getIsPlaceableOverLand() or some such?

Thoughts?

Hamms: Alternatively, we might want to just use `isFlat` for agent, since the only functional restriction on what it should be able to place is stuff that doesn't require a "hop off" animation

WilliamHarmonMS: That's a good suggestion, but I wondered if that might allow for the placement of too many blocks? Are there -no- other blocks other than redstone wire / pressure plates/ rails that would be included in isFlat?

I mean I guess it should be able to place regular torches if it can place redstone torches? What about stuff like flowers? Those are always walkable, right? Are they "flat"?

Hamms: [not flat, no](https://github.com/code-dot-org/craft/blob/master/src/js/game/LevelMVC/LevelBlock.js#L325-L329)

WilliamHarmonMS:
@joshlory or @Hamms 
Whenever you guys have time, addressed the isFlat() feedback :)

Hamms:
We actually probably want to update `isFlat` to consider `isRail` rather than `blockType.substring(0, 5) === "rails"`, since torches aren't actually flat, and then actually look at `isWalkable` for the placement check since that's [_actually_ what we care about](https://github.com/code-dot-org/craft/blob/master/src/js/game/LevelMVC/LevelView.js#L980) (I got that one wrong in my comment, sorry for the misdirect)

src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -323,9 +324,11 @@ module.exports = class LevelBlock {
    * @return {boolean}
    */
   static isFlat(blockType) {
-    return blockType.substring(0, 5) === "rails" ||
+    // Starts With 'rail' will capture rails and railsRedstoneTorch.
+    return blockType.isRail ||
         blockType.startsWith("redstoneWire") ||
-        blockType.startsWith("pressurePlate");
+        blockType.startsWith("pressurePlate") ||
+        blockType === "torch";
```
Hamms: we still don't want torches to be `flat`

WilliamHarmonMS: ack, I put that before our conversation and hard just forgot to take it out.

PR #340, opened 2017-10-19
Don't apply the tint variation effect to minecart rails

joshlory:
Now that tint variation is applied on a per-block basis, we can choose which block types receive tint and which don't.

Before: Rails receive shadows, but are darkened by being below the `Shade_Layer.png`.

darker

After: Rails still receive shadows, but are as bright as before.

brighter

See:
 - PR https://github.com/code-dot-org/craft/pull/279: Render "flat" blocks in Ground Plane
 - PR https://github.com/code-dot-org/craft/pull/293: Replace shade layer with tint


src/js/game/LevelMVC/LevelView.js
```diff
@@ -2187,7 +2187,9 @@ module.exports = class LevelView {
         yOffset = this.blocks[blockType][3];
         sprite = group.create(xOffset + 40 * x, yOffset + group.yOffset + 40 * y, atlas, frame);
         if (group === this.actionGroup || group === this.groundGroup) {
-          this.psuedoRandomTint(group, sprite, x, y);
+          if (!blockType.startsWith("rails")) {
```
Hamms: this should probably be checking `.isRails`, unless you want this to also apply for `railsRedstoneTorch`

PR #345, opened 2017-10-19
just picking up items on rails

WilliamHarmonMS:
@joshlory 
This is, I think, JUST the collecting items while on a rail. Since the 'use' rail riding has been taken out, it's a bit hard to test this at the moment, but [PR#332](https://github.com/code-dot-org/craft/pull/332) fixes that.

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -764,7 +764,6 @@ module.exports = class LevelView {
     } else {
       this.onAnimationEnd(this.playMinecartMoveForwardAnimation(position, facing, isOnBlock, completionHandler, nextPosition, speed), () => {
         this.playTrack(nextPosition, nextFacing, isOnBlock, entity, completionHandler);
-        entity.collectItems(entity.position);
       });
     }
   }
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -1923,6 +1923,7 @@ module.exports = class LevelView {
     sprite = this.actionGroup.create(xOffset + 40 * x, yOffset + 40 * y, atlas, "");
     const anim = sprite.animations.add("animate", frameList, 10, false);
 
+    // If direct player control, we have stuff to do to manage miniblock pick up
     if (this.controller.getIsDirectPlayerControl()) {
       const distanceBetween = function (position, position2) {
         return Math.sqrt(Math.pow(position[0] - position2[0], 2) + Math.pow(position[1] - position2[1], 2));
@@ -1933,12 +1934,11 @@ module.exports = class LevelView {
         if (this.controller.levelModel.usePlayer) {
           if (distanceBetween(this.player.position, collectiblePosition) < collectibleDistance) {
             this.playItemAcquireAnimation(this.player.position, this.player.facing, sprite, () => { }, blockType);
-          } else {
-            this.collectibleItems.push([sprite, [xOffset, yOffset], blockType, collectibleDistance]);
           }
         }
       });
     }
+    this.collectibleItems.push([sprite, [xOffset, yOffset], blockType, collectibleDistance]);
     this.playScaledSpeed(sprite.animations, "animate");
     return sprite;
   }
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -1930,15 +1930,16 @@ module.exports = class LevelView {
       };
 
       const collectiblePosition = this.controller.levelModel.spritePositionToIndex([xOffset, yOffset], [sprite.x, sprite.y]);
+      this.collectibleItems.push([sprite, [xOffset, yOffset], blockType, collectibleDistance]);
       anim.onComplete.add(() => {
         if (this.controller.levelModel.usePlayer) {
           if (distanceBetween(this.player.position, collectiblePosition) < collectibleDistance) {
-            this.playItemAcquireAnimation(this.player.position, this.player.facing, sprite, () => { }, blockType);
+            this.player.collectItems([x,y]);
           }
         }
       });
     }
-    this.collectibleItems.push([sprite, [xOffset, yOffset], blockType, collectibleDistance]);
+
     this.playScaledSpeed(sprite.animations, "animate");
     return sprite;
   }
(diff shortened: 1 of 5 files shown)


src/js/game/LevelMVC/LevelView.js
```diff
@@ -764,6 +764,7 @@ module.exports = class LevelView {
     } else {
       this.onAnimationEnd(this.playMinecartMoveForwardAnimation(position, facing, isOnBlock, completionHandler, nextPosition, speed), () => {
         this.playTrack(nextPosition, nextFacing, isOnBlock, entity, completionHandler);
+        entity.collectItems(entity.position);
```
Hamms: I'm not sure I love the idea of indiscriminately calling a method that's only defined for the Player entity here, much less the idea of calling a method that was formerly functionally private. Is there somewhere on the Player entity itself we could call this? Perhaps update the player position as it moves around the track, and have that trigger item collection?

WilliamHarmonMS: position is already updated while going around a track so updateMovement() would be a perfectly viable place to put this.

I'm torn on that though, because having a if(onTracks) conditional incorporated into every updateMovement() call, on the one hand, doesn't sound... logical? If we know we want the pick up and track riding to be linked, calling the pick up in the track ride logic makes the most sense to me. 

The easy trap there is, you're accurate, that even though that's 100% safe at the moment (since the only thing that can ride rails is the player) we don't TECHNICALLY know that that will always be the case, so that method isn't very future-proof. If an Agent could ever ride a rail, then we wouldn't want it to pick up miniblocks and then that function call could cause a problem. In the end, I'm not crazy invested either way, so I can move it to updateMovement on Player.js 👍

WilliamHarmonMS:
@Hamms PTAL and see if that placement looks better to you :)

Hamms:
I was actually thinking of a place we could unite the "on the rails" collect and the "regular movement" collect. Maybe something like [updating LevelModel here](https://github.com/code-dot-org/craft/blob/master/src/js/game/LevelMVC/LevelModel.js#L792) to instead call an `entity.setPosition` method, which can in player have a side effect of triggering a collect check

WilliamHarmonMS:
@Hamms Let me know if this looks better 👍 
PTAL

WilliamHarmonMS:
@Hamms Removed the other this.collectItems references and everything is looking to still work. (tested by walking by miniblocks generated through block destruction as well as diamondMiniblocks manually put into the level). 

If unit tests are a blocker for this PR what would be the best way to test this do you think? Is there a way to query how many active mini blocks are in the level? Walking by a diamondMiniblock updates the levelModel.actionPlane to reflect a then empty block? Or no?

WilliamHarmonMS:
@joshlory or @Hamms 
Added a unit test to this and it seems to be all green and working as expected 👍

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1933,12 +1934,11 @@ module.exports = class LevelView {
         if (this.controller.levelModel.usePlayer) {
           if (distanceBetween(this.player.position, collectiblePosition) < collectibleDistance) {
             this.playItemAcquireAnimation(this.player.position, this.player.facing, sprite, () => { }, blockType);
-          } else {
-            this.collectibleItems.push([sprite, [xOffset, yOffset], blockType, collectibleDistance]);
           }
         }
       });
     }
+    this.collectibleItems.push([sprite, [xOffset, yOffset], blockType, collectibleDistance]);
```
Hamms: Why did this get moved down here?

WilliamHarmonMS: @Hamms 
I spoke to Josh this morning about this. This was initiated by the fact that the unit tests weren't picking it up because we weren't mocking enough data, but I ended up moving it because before it was only getting tracked in the collectibleItems array if both:
this.controller.getIsDirectPlayerControl() && this.controller.levelModel.usePlayer

The problem was that even if those checks failed, the miniblock is still created, it's just not tracked anywhere and thus lost to the void. I moved the array tracking to the bottom of the function so that regardless of whether or not we have a player that's picking them up, we still keep track of them. 

This had a side effect of making the unit test easier, so no complaints on that front either. 👍

Hamms: Won't this have the side effect of also tracking collectibles that are collected by the player immediately upon spawning?

WilliamHarmonMS: I tested that and it looked like the collectibleItems array was being updated correctly. The unit test has the player at [0,0] and spawns miniblocks at [0-5, 0], and then has the player walk right 3 times. The result should be 3 mini blocks picked up, and 3 mini blocks left behind (with one of the miniblocks being picked up immediately, since player at [0,0] and a block spawned at [0,0]).

The result of the unit test is that there are 3 remaining element in the collectibleItems array, as expected.

It seems to be working?

Hamms: right, but the item at [0, 0] probably isn't getting picked up until the player moves, when it should get picked up right away.

WilliamHarmonMS: It does get picked up right away still, I think the risk is that it will remain in the collectibleItems array until the player moves... That's a fair point, if we consider that a concern? Though if it's cleaned up on moving, and that's only a problem in levels where the player would be manually controlling the player, perhaps it's low priority? 

What do you think?

Hamms: We definitely consider that a concern for all the 2015 levels

WilliamHarmonMS: Well than it shall be fixed! It was more a change to appease the unit test anyway, so any actual functional concerns definitely take precedence.

WilliamHarmonMS:
@Hamms 

So actually found a system that I like the logic of that shouldn't effect previous years and still maintains unit test success status.

basically: if a miniblock is created, ever, collectibleItems array becomes aware of it. The problem with the previous iteration was that the automatic clean up if the player was close enough wasn't addressing that array. It just played the pick up animation and nothing else. This has been solved by switching from a direct animation call:
this.playItemAcquireAnimation()
to an entity driven pick up call:
this.player.collectItems()

the entity driven pick up does more to manage the state of not only the player's inventory(??? because this.player.inventory is apparently a thing?), but also the list of collectibleItems on the level. Quick tests show that now, when you are close enough to pick a block up immediately, the collectibleItems array correctly reflects the appropriate number of items at all times.

Let me know what you think 😄

WilliamHarmonMS:
@Hamms or @joshlory 

PR #347, opened 2017-10-19
add craftCollectibleCollected event

Hamms:
In order to track the Diamond progression path, we need a way to hook in to the collection event.

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -1,5 +1,6 @@
 const LevelBlock = require("./LevelBlock.js");
 const FacingDirection = require("./FacingDirection.js");
+const createEvent = require("../../utils").createEvent;
 
 module.exports = class LevelView {
   constructor(controller) {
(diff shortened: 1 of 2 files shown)


Hamms:
This is one such way we could go about doing what we need to do, but it is a super-specific approach. Another approach that's a bit less specialized would be to have a more generic set of events on the GameController that other entities could hook in to (similar to [what we do for Blockly](https://github.com/code-dot-org/blockly/blob/staging/core/ui/block_space/block_space.js#L82)).

@joshlory thoughts on this approach vs the more extensible one?

joshlory:
This approach looks fine.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1374,6 +1375,9 @@ module.exports = class LevelView {
           (this.player.inventory[blockType] || 0) + 1;
         sprite.kill();
         this.toDestroy.push(sprite);
+        const e = createEvent('craftCollectibleCollected');
+        e.blockType = blockType;
+        window.dispatchEvent(e);
```
joshlory: Tiny nit: `event`

PR #353, opened 2017-10-20
Water & lava ambient occlusion

joshlory:
Before:

lava-glow-before

After:

lava-glow-after


joshlory:
Before:

water-ao-before

After:

water-ao-after

src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -300,7 +300,7 @@ module.exports = class LevelBlock {
   }
 
   getIsEmptyOrEntity() {
-    return this.isEmpty || this.isEntity;
+    return this.isEmpty || this.isEntity || this.getIsLiquid();
```
Hamms: It feels like we're violating the nature of this check. Could we instead add a method like `getCanOcclude` that uses `getIsEmptyOrEntity` and `getIsLiquid`?

joshlory: Good call, I will fix!

WilliamHarmonMS:
Comment from Joel:
"They look good except for one thing. In the lava it looks like some action plane blocks are casting AO over the lava and shouldn't."

I think this is in reference to the right side of the lava block when it's to the left of a block in the actionPlane? Attaching image.
screen shot 2017-10-23 at 10 25 19 am

joshlory:
Yep, Joel is 100% right!  I fixed the edge-of-the-world AO shadow on lava, but AO from blocks on the action plane is a bit harder.  I may address it in this PR, or possibly a follow-up PR (I don't think this blocks shipping since this is already an issue in the "Before" version).

One other issue I see is that block shadows on water are a few px too high.  This is an existing issue, but is maybe a little more obvious with the new AO changes:

image

Is this worth addressing?  I think we'd just need a small 4px parallelogram shadow for the ground plane block.

joshlory:
Additionally, block shadows continue to cast on lava: (existing issue)

image

joshlory:
@Hamms @WilliamHarmonMS PTAL!

PR #356, opened 2017-10-20
New nether portal

WilliamHarmonMS:
new Nether Portal sprites to be used with transparent action blocks.


joshlory:
Thinking about this more, how do tree leaves on the fluff plane work as a separate asset?

https://github.com/code-dot-org/craft/blob/master/src/assets/images/Leaves_Spruce_Decay.png

WilliamHarmonMS:
@joshlory 
In levelView.createBlock() there is a locally defined anonymous function assigned to buildTree:

var buildTree = function (levelView, frame)

that looks like it creates a fluff plane block that then animates when the actionPlane block (the "tree" trunk) is destroyed. I looked into this when first looking at the Nether portal, but it didn't look like we could hijack it 1:1 because of some of the differences between how the two behave (one animates all the time, vs on destruction. One occupies two actionPlane indices vs just one... so on).

it would likely be possible to utilize a similar method of placing "Nether_Portal" on the action plane, and then populating the portal itself in the fluff plane, if we wanted, but it's such a 1-time thing, I didn't figure the system required extensive robustness? Let me know what you think.

joshlory:
My preference would be to follow the existing pattern for tree leaves, if possible.  Not necessarily using the same function, but using the same approach of placing the top part of the portal in the fluff plane.

This also makes it easier to implement fading opacity of the portal when an entity walks behind it.

WilliamHarmonMS:
I can look into that today. Is that a blocker for this PR? This was merely adding a new .png file to the assets directory. I feel like altering the functionality of the Nether_Portal block objects might be different enough to be its own PR?

joshlory:
The main issue is that right how we have some caching issues around changing both the block.png and blocks.json 😕.  I'd like to avoid churn in those files if possible!

WilliamHarmonMS:
ahhhh, I do remember... Ryan? Saying something about that a while ago? I can hold off on this PR if it would cause problems for any classroom tests this week. Just let me know :)

WilliamHarmonMS:
@Hamms or @joshlory 

Issue #105, opened 2017-09-13
Rails should "forget" old connections when the connecting rail is destroyed


joshlory:
### Repro

1. Create a curved track:

![screen shot 2017-09-13 at 10 53 10 am](https://user-images.githubusercontent.com/413693/30392459-c2414df2-9871-11e7-8e6d-b7cf956002f9.png)

2. Destroy one of the connected rails:

![screen shot 2017-09-13 at 10 54 37 am](https://user-images.githubusercontent.com/413693/30392533-f7c4bbbc-9871-11e7-976b-a5b69ddca81c.png)

3. Place a track to create a straight line.  The curved track should straighten out.  Instead it remembers the old `connectionA` or `connectionB` and doesn't re-form a horizontal track.

### Expected

![screen shot 2017-09-13 at 10 50 06 am](https://user-images.githubusercontent.com/413693/30392470-d0de9608-9871-11e7-9bcf-52783ae9b6ac.png)

### Actual

![screen shot 2017-09-13 at 10 50 39 am](https://user-images.githubusercontent.com/413693/30392430-ade497ba-9871-11e7-9a62-aae34901fd58.png)

PR #357, opened 2017-10-21
When rail blocks are removed, sever the connections from adjacent rail blocks

Hamms:
This allows for newly placed rail blocks to "heal" rails whose
connections were removed, while still not updating rails that are
already happily connected

Fixes https://github.com/code-dot-org/craft/issues/105


joshlory:
Fantastic cleanup!

It looks like the Travis CI failure may be legit.

Hamms:
Ah, yep; a bit of my WIP powered rails change snuck in. Rebased to remove it

PR #364, opened 2017-10-23
if a sheep is naked, drop() will now exit out before calling super.dr…

WilliamHarmonMS:
…op()

@joshlory exiting out of the sheep's drop command if the sheep is naked, so we don't ever hit super.drop(), or any subsequent functionality.

### DIFF
--- a/src/js/game/Entities/Sheep.js
+++ b/src/js/game/Entities/Sheep.js
@@ -2,11 +2,11 @@ const BaseEntity = require("./BaseEntity.js");
 const EventType = require("../Event/EventType.js");
 
 module.exports = class Sheep extends BaseEntity {
-    constructor(controller, type, identifier, x, y, facing, skipView = false) {
+    constructor(controller, type, identifier, x, y, facing) {
         super(controller, type, identifier, x, y, facing);
         var zOrderYIndex = this.position[1];
         this.offset = [-43, -55];
-        if (!skipView) {
+        if (this.controller.levelView) {
           this.prepareSprite();
           this.sprite.sortOrder = this.controller.levelView.yToIndex(zOrderYIndex);
         }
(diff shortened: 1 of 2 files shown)


joshlory:
Test coverage? 😄

WilliamHarmonMS:
@joshlory test coverage added 👍

src/js/game/Entities/Sheep.js
```diff
@@ -2,12 +2,14 @@ const BaseEntity = require("./BaseEntity.js");
 const EventType = require("../Event/EventType.js");
 
 module.exports = class Sheep extends BaseEntity {
-    constructor(controller, type, identifier, x, y, facing) {
+    constructor(controller, type, identifier, x, y, facing, skipView = false) {
```
joshlory: Rather than passing in this new param, can we skip `prepareSprite()` when `this.controller.levelView` is undefined?

test/unit/EntityTest.js
```diff
@@ -70,3 +71,16 @@ test('canPlaceBlockOver', t => {
   t.end();
 });
 
+test('sheep Drop', t => {
+  const sheep = new Sheep(mockGameController, "Sheep", 1, 1, "Sheep", true, 1);
+
+  // Sheep starts !naked
+  t.false(sheep.naked);
```
joshlory: I wonder if we should change this property to `sheared` instead.

joshlory:
LGTM after addressing the `skipView = false`.

src/js/game/Entities/Sheep.js
WilliamHarmonMS: 👌

test/unit/EntityTest.js
WilliamHarmonMS: What if a sheep is naked because it just doesn't grow fur (hair? wool? Is it still called wool, pre-shearing?) and not because it was sheared!?

Either makes sense to me, if anyone some day is passionate about it, it could be changed with little effort.

PR #365, opened 2017-10-23
if on an agent level, 'use' command simply exits out

WilliamHarmonMS:
@joshlory 

if we're in an agent level, the 'use' command should be exited out of. This feature was requested this morning as the direction people want to go with in regards to that functionality.

### DIFF
--- a/src/js/game/GameController.js
+++ b/src/js/game/GameController.js
@@ -990,6 +990,10 @@ class GameController {
   }
 
   use(commandQueueItem) {
+    if (this.levelData.isAgentLevel) {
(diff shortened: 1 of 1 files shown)


joshlory:
We should talk this over a bit with Ali & Cole.  When testing the auto-rails branch I found the lack of a "use" to start riding in a minecart to be pretty jarring.  It might work on an enclosed level, but I think it'll be frustrating for users on a free play level.

joshlory:
Confirmed with @ryansloan that we don't want to teach the "use" button for 2017.

Issue #358, opened 2017-10-23
The Agent destroys a diamond when placing a rail on top of it.


v-jaarba:
Found in: https://levelbuilder-studio.code.org/s/craft17/stage/1/puzzle/12

When the Agent is building rails if they place the rail on a location with a diamond the diamond is destroyed. Not in parity with Minecraft. The agent does not seem to be collecting the diamonds. This could cause issue if we ever require the collection of diamonds in the future.

**Repro Steps:**
1) Start level 12 in HoC.
2) Make the Agent place rails over diamonds.
3) Observe.

**Expected Results:**
The diamonds are not destroyed by placing rails. 

**Actual Results:**
The diamonds are destroyed by placing rails.

**Video:**
[DestroyDiamonds.zip](https://github.com/code-dot-org/craft/files/1408033/DestroyDiamonds.zip)

v-jaarba:
Any place item, placed on top of collectable destroys the collectable.

v-jaarba:
This issue is still occurring in level 10, place a rail on any mini block, pressure plate, torch or win condition item the rail will destroy the item.

Issue #360, opened 2017-10-23
The Player can not finish the level if they are in a minecart when they enter the win condition location.


v-jaarba:
Found in: https://levelbuilder-studio.code.org/s/craft17/stage/1/puzzle/10

The player could create a situation where they use the mine cart to enter the win condition location.
If they do the player loses control of Alex/Steve and will have to reset the level to continue.

**Repro Steps:**
1) Start level 10. 
2) Make the agent place rails on top of the win condition location.
3) Have Steve/Alex use the mine cart to move into the win condition location.
4) Observe.

**Expected Results:**
Either the player wins the level and Steve/Alex automatically get out of the mine cart. Or the player can continue to play and get out of the mine cart and can still maneuver Steve/Alex to win.

**Actual Results:**
Player loses control of Steve/Alex and the level can no longer be completed.

**Video:**
[MinecartFTW.zip](https://github.com/code-dot-org/craft/files/1408109/MinecartFTW.zip)

PR #366, opened 2017-10-23
placing blocks over a diamondMiniblock would destroy the diamond

WilliamHarmonMS:
@joshlory 

this fixes [issue#358](https://github.com/code-dot-org/craft/issues/358)


WilliamHarmonMS:
This actually ALSO happens to fix [issue#360](https://github.com/code-dot-org/craft/issues/360) as all solution indices have miniBlocks in them.

Technically however, if we wanted to future proof this, the fix could be extended to cover 360 in a more robust fashion.

PR #369, opened 2017-10-23
adding an invisible block for usre with NetherPortal

WilliamHarmonMS:
@joshlory 

An invisible block will get shadows and collision behavior. These will be used for the base of the NetherPortal.

### DIFF
--- a/src/js/game/LevelMVC/LevelBlock.js
+++ b/src/js/game/LevelMVC/LevelBlock.js
@@ -77,6 +77,7 @@ module.exports = class LevelBlock {
       this.isDestroyable = false;
       this.isUsable = false;
       this.isWeaklyPowerable = false;
+      this.isEmissive = true;
     }
 
     if (blockType.startsWith("glass")) {
--- a/src/js/game/LevelMVC/LevelBlock.js
+++ b/src/js/game/LevelMVC/LevelBlock.js
@@ -73,6 +73,12 @@ module.exports = class LevelBlock {
       this.isUsable = true;
     }
 
+    if (blockType === "invisible") {
+      this.isDestroyable = false;
+      this.isUsable = false;
+      this.isWeaklyPowerable = false;
+    }
+
     if (blockType.startsWith("glass")) {
       this.isSolid = false;
     }
--- a/src/assets/images/Blocks.json
+++ b/src/assets/images/Blocks.json
@@ -1646,6 +1646,15 @@
 	"spriteSourceSize": {"x":13,"y":2,"w":40,"h":60},
 	"sourceSize": {"w":64,"h":64},
 	"pivot": {"x":0.5,"y":0.5}
+},
+"Invisible":
+{
+	"frame": {"x":511,"y":0,"w":1,"h":1},
+	"rotated": false,
+	"trimmed": true,
+	"spriteSourceSize": {"x":511,"y":0,"w":1,"h":1},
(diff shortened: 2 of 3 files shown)


Hamms:
Yeah, I'd very much like to have a reliable approach that adds a dedicated transparency pixel

WilliamHarmonMS:
Roger that, I'll ping the man and let him know.

joshlory:
Like we talked through in https://github.com/code-dot-org/craft/pull/356#issuecomment-338348611, can we have the top half of the portal live in the fluff plane like tree leaves?

That will be more consistent with other block decorations, and will get rid of the need for a 1x1 transparent pixel block.

WilliamHarmonMS:
We would need some block to exist in the actionPlane so that we get collision / shadows at the base indices. So presuming the fluff plane is drawn on top of the action plane, we'd still need to put down something that's unbreakable and blocks movement... so bedrock? It feels weird to me to use bedrock as the base, as opposed to a block that is explicitly a filler just to get collision / shadows? Unless we want to write new logic into move commands that say "is this index a NetherPortal OR is this index 4 indices to the right of a NetherPortal (to block movement into the other base index)? Then don't move", but that feels -really bad- to me.

Also, I spoke with Joel about the invisible pixel in the JSON, and he's made another file so if future changes are made to the .png, the invisible block will be included, so there shouldn't be any problems on that front.

joshlory:
I was thinking the base blocks would be obsidian, like in the original design.

WilliamHarmonMS:
Update on this: Joel had the request that we make these invisible blocks clear FoW (similar to torches) and that would give the desired effect for the NetherPortal emitting light correctly (given that they're only 2 indices apart). However, additionally, he's changed the name from Invisible to netherPortalBase to be more explicit that they wouldn't be general all purpose filler blocks (because if they clear FoW, they're really only appropriate to be used with the nether portal, not as an all purpose for all situations). 

Having both base blocks of the Nether Portal clear FoW to simulate the effect the nether portal has in Minecraft (since it does indeed emit light), would, I think, disincentivize the use of obsidian?

WilliamHarmonMS:
@Hamms or @joshlory 

joshlory:
For invisible blocks, can we just not add a sprite to the `actionGroup`?  Or does this get messy when the sprite is destroyed?

WilliamHarmonMS:
@joshlory 

I haven't looked into that yet. I would hope that that would be a fine solution, but given how many references to model.player were assumed when we started looking into the Agent stuff, I'd be staggered if there weren't -also- a bunch of sprite clean up logic that assumes there is a sprite to clean up.

I can look into this today if there is time, other wise this PR does completely solve the problem (Though I might need to go in and make these blocks emit light / clear FoW? I forget if I addressed that already or not. Haven't checked this branch since last week I don't think)

src/js/game/LevelMVC/LevelBlock.js
```diff
@@ -73,6 +73,12 @@ module.exports = class LevelBlock {
       this.isUsable = true;
     }
 
+    if (blockType === "invisible") {
+      this.isDestroyable = false;
+      this.isUsable = false;
+      this.isWeaklyPowerable = false;
```
joshlory: To have this block clear FoW I think you can set `this.isEmissive = true;`

WilliamHarmonMS: 👌

Issue #367, opened 2017-10-23
Using place on top of rails will 'destroy' the rail. 


v-jaarba:
Find in: https://levelbuilder-studio.code.org/s/craft17/stage/1/puzzle/10

When using the place block onto a rail that is already in level will force the rail to disappear.
This is not the destroy functionality of the destroy block as that pops the rail off for collection.
This just removes the rail from the level.

**Repro Steps:**
1) Start level 10.
2) Have the agent move over a rail in the level and have it place anything down that is not a rail.
3) Observe.

**Expected Results:**
The rails are only destroyed if the place block can place something in that location. It should also pop the rail off like the destroy block.

**Actual Results:**
The place block will remove the rail block from the location and nothing else will be placed.

**Video:**
[Place-replace.zip](https://github.com/code-dot-org/craft/files/1408822/Place-replace.zip)

PR #371, opened 2017-10-24
trying to place inappropriate blocks would still destroy rails

WilliamHarmonMS:
This fixes [issue#367](https://github.com/code-dot-org/craft/issues/367)

Cannot BELIEVE that levelModel.canPlaceBlock() just returned true 😂

### DIFF
--- a/src/js/game/GameController.js
+++ b/src/js/game/GameController.js
@@ -1186,9 +1186,10 @@ class GameController {
   placeBlock(commandQueueItem, blockType) {
     const player = this.getEntity(commandQueueItem.target);
     const position = player.position;
-    let blockTypeAtPosition = this.levelModel.actionPlane.getBlockAt(position).blockType;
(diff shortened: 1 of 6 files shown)


WilliamHarmonMS:
@joshlory or @Hamms 
didn't ping either of you in the initial comment. 😂 
PTAL

src/js/game/GameController.js
```diff
@@ -1227,7 +1228,7 @@ class GameController {
         }, this);
       }
     } else {
-      commandQueueItem.failed();
+      commandQueueItem.succeeded();
```
Hamms: why succeed in this case?

WilliamHarmonMS: because failed causes you to lose control of your entity, and this was simply NEVER hit because the if case that corresponds with that else{} was this.levelModel.canPlaceBlock(), which simply did nothing but return true;

🤢

PR #375, opened 2017-10-24
desynch ghast animations

WilliamHarmonMS:
@josh or @Hamms 

randomized the order of the ghast idle animation so that multiple ghasts on the same level won't be on the same frame at the same time.


src/js/game/Entities/Ghast.js
```diff
@@ -18,7 +18,7 @@ module.exports = class Ghast extends BaseEntity {
         var frameName = "Ghast";
         this.sprite = actionGroup.create(0, 0, 'ghast', 'Ghast0000.png');
         this.sprite.scale.setTo(1,1);
-        let stillFrameName = ['Ghast0072.png', 'Ghast0048.png', 'Ghast0024.png', 'Ghast0000.png'];
+        //let stillFrameName = ['Ghast0072.png', 'Ghast0048.png', 'Ghast0024.png', 'Ghast0000.png'];
```
joshlory: Not needed?

src/js/game/Entities/Ghast.js
```diff
@@ -30,8 +30,19 @@ module.exports = class Ghast extends BaseEntity {
 
             // idle sequence
             frameList = Phaser.Animation.generateFrameNames(frameName, frameListPerDirection[i][0][0], frameListPerDirection[i][0][1], ".png", 4);
+
+            let randomOffset = Math.floor(Math.random() * (frameList.length - 2)) + 2;
```
joshlory: Let's extract this as a `randomInt(min, max)` helper in utils.

joshlory:
@WilliamHarmonMS Can this PR be updated to also use the `setTimeout` delay, instead of adding duplicate frames?

I don't think it's as urgent since there won't be nearly as many Ghast entities on screen at once.  I'm ok with merging this now and fixing it up later (if you open an issue to track removing extra frames in entity animations and replacing with `setTimeout`).

WilliamHarmonMS:
@joshlory Will do!

opening an issue now, and will merge this one.

[issue:399](https://github.com/code-dot-org/craft/issues/399)

PR #376, opened 2017-10-24
Add Position helper

Hamms:
Similar to FacingDirection, this class is right now a new place to add common positioning logic helper methods like equivalency and movement that are right now scattered throughout our system.

It's also implemented as a class, which should make it much easier in the future to make the transition from simple `[x, y]` arrays to actual Position instances.

### DIFF
--- a/src/js/game/LevelMVC/FacingDirection.js
+++ b/src/js/game/LevelMVC/FacingDirection.js
@@ -33,6 +33,15 @@ const FacingDirection = Object.freeze({
       case FacingDirection.East: return [1, 0];
       case FacingDirection.West: return [-1, 0];
     }
+  },
+
+  directionToRelative(direction) {
+    switch (direction) {
+      case FacingDirection.North: return "Up";
+      case FacingDirection.South: return "Down";
+      case FacingDirection.East: return "Right";
+      case FacingDirection.West: return "Left";
+    }
   }
 });
 
--- a/src/js/game/LevelMVC/LevelPlane.js
+++ b/src/js/game/LevelMVC/LevelPlane.js
@@ -10,12 +10,14 @@ const {
   directionToOffset
 } = require("./FacingDirection.js");
 
+const Position = require("./Position");
+
 const connectionName = function (connection) {
   switch (connection) {
-    case 0: return 'North';
-    case 2: return 'South';
-    case 1: return 'East';
-    case 3: return 'West';
+    case North: return 'North';
+    case South: return 'South';
+    case East: return 'East';
+    case West: return 'West';
     default: return '';
   }
 };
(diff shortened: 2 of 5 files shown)


src/js/game/LevelMVC/FacingDirection.js
```diff
@@ -33,6 +33,15 @@ const FacingDirection = Object.freeze({
...
       case FacingDirection.West: return [-1, 0];
     }
+  },
+
+  directionToRelative(direction) {
+    switch (direction) {
+      case FacingDirection.North: return "Up";
+      case FacingDirection.South: return "Down";
+      case FacingDirection.East: return "Right";
+      case FacingDirection.West: return "Left";
```
joshlory: Should these be "North", "South", "East", "West" for consistency?  (Updating the piston block names to match.)

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -7,15 +7,18 @@ const {
...
 const connectionName = function (connection) {
   switch (connection) {
-    case 0: return 'North';
-    case 2: return 'South';
-    case 1: return 'East';
-    case 3: return 'West';
+    case North: return 'North';
+    case South: return 'South';
+    case East: return 'East';
+    case West: return 'West';
```
joshlory: Could be replaced by `directionToRelative` if we make the change above.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -167,9 +170,9 @@ module.exports = class LevelPlane {
       this.determineRailType(position, true);
 
       if (this.levelModel && this.levelModel.controller.levelView) {
-        const northEast = [position[0] + 1, position[1] - 1];
-        const southWest = [position[0] - 1, position[1] + 1];
-        let positionAndTouching = this.getOrthogonalPositions(position).concat([position, northEast, southWest]);
+        const northEast = Position.add(position, [1, -1]);
+        const southWest = Position.add(position, [-1, 1]);
```
joshlory: Worth adding `position.south().west()`?

Until everything is a `Position` instance it would be `Position.west(Position.south(position))` 🙁.

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -257,7 +236,7 @@ module.exports = class LevelPlane {
     const speed = 300;
 
     if (block.connectionA === facing || block.connectionB === facing) {
-      return ["", this.forwardPosition(position, facing), facing, speed];
+      return ["", Position.forward(position, facing), facing, speed];
```
joshlory: 👍

src/js/game/LevelMVC/Position.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  static equals(left, right) {
+    return left[0] === right[0] && left[1] === right[1];
+  }
+
+  static isAdjacent(left, right) {
+    return directions
+        .map(FacingDirection.directionToOffset)
+        .some(offset => Position.equals(Position.add(left, offset), right));
```
joshlory: Tiny style nit: I've been doing line continuations at 2 space indent in this repo, even though we're not consistent about it in the main `code-dot-org` repo.

joshlory:
Nice cleanup!

src/js/game/LevelMVC/FacingDirection.js
Hamms: I can see the value in updating pistons to use cardinal rather than relative directions, but it's _definitely_ out of scope for this PR

joshlory: Ah right, because it would require changes to the levels as well 😕.  Sounds good!

PR #378, opened 2017-10-24
Add Powered rails

Hamms:
This PR does a bit more work than it needs to do, but it's laying a fair amount of groundwork for the redstone refactor I'm working on.

Notably, it [adds the concept of Adjacency Sets](https://github.com/code-dot-org/craft/commit/2302148aff5b09b740bdbd0dd51e83be34bb6167) to the Position Helper. This specific implementation of adjacency sets should be considered at least partially a proof of concept; it's significantly more computationally complex than it should be, and I intend to improve that.

With that done, the [addition of a `powerRails` step](https://github.com/code-dot-org/craft/commit/71797c056ab3ebe8cb29207b735504b650459023) simply uses a custom definition of "adjacency" to gather connectivity sets of rails.


src/js/game/LevelMVC/LevelModel.js
```diff
@@ -357,19 +357,6 @@ module.exports = class LevelModel {
...
-      for (var y = 0; y < this.planeHeight; ++y) {
-        var block = this.actionPlane.getBlockAt([x, y]);
-        if (block.blockType.substring(0, 7) === "railsUn") {
-          unpoweredRails.push([x, y], "railsPowered" + block.blockType.substring(14));
-        }
-      }
-    }
-    return unpoweredRails;
-  }
-
```
joshlory: 🎉

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -253,6 +253,86 @@ module.exports = class LevelPlane {
...
   /**
+   * Determine whether or not the blocks at the given positions are powered
+   * rails that are connected to each other
+   *
+   * @param {Posititon} left
+   * @param {Posititon} right
+   * @return {boolean}
+   */
+  getPoweredRailsConnected(left, right) {
+    // return early if they're not even adjacent
```
joshlory: Nit: `// Return early if the positions are not event adjacent.`

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -253,6 +253,86 @@ module.exports = class LevelPlane {
   }
 
   /**
+   * Determine whether or not the blocks at the given positions are powered
+   * rails that are connected to each other
```
joshlory: Tiny nit: missing punctuation.

joshlory:
I think this change may regress https://github.com/code-dot-org/craft/issues/105?  Specifically, this (skipped) test passes on master but fails on this branch: https://github.com/code-dot-org/craft/blob/807a39627c36f384c1b33edd0441d57af6618b67/test/unit/LevelPlaneTest.js#L245

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -253,6 +253,86 @@ module.exports = class LevelPlane {
...
+    // to be connected, both blocks must be powerable rails
+    if (!(leftBlock.getIsPowerableRail() && rightBlock.getIsPowerableRail())) {
+      return false;
+    }
+
+    // to be connected, both blocks must be oriented either North/South or
+    // East/West
+    const bothEastWest =
+      leftBlock.blockType.match('East|West') &&
+      rightBlock.blockType.match('East|West');
```
joshlory: Worth adding a `rightBlock.isHorizontal()` helper method?  A bit gross now when all block types live in the same class, but might make sense in the future when the different block types are separate classes.

Hamms:
Huh; why is that test getting skipped?

joshlory:
Never mind!  I had pulled a local `powered-rails` branch that happened to be named the same as your PR branch 😄.

Hamms:
That test is passing on this branch for me ... can you confirm your repro?

src/js/game/LevelMVC/LevelPlane.js
```diff
@@ -253,6 +253,86 @@ module.exports = class LevelPlane {
...
+      return false;
+    }
+  }
+
+  /**
+   * Propagate power to (and orient) all powerable rails in the level.
+   */
+  powerRails() {
+    // find all rails that can be powered
+    const powerableRails = this.getAllPositions().filter((position) => (
```
joshlory: Tiny nit: extra parens around `position`.

PR #384, opened 2017-10-26
Sparkle

WilliamHarmonMS:
@Hamms 

Redstone sparkles will:
(1) appear over charged redstone
(2) play 1 of the 3 sparkle animations
(3) loop
(4) have a random delay between loops
(5) play on a random offset on each new loop
(6) disappear when the redstone is uncharged

This has been signed off by Joel and is looking real nice!


joshlory:
Hmm, some of the sparkle animations seem to be too far removed from the actual wires?

image

joshlory:
I'm seeing a stale sprite after destroying the redstone torch in the default level:

image

src/js/game/LevelMVC/LevelView.js
```diff
@@ -2096,28 +2099,69 @@ module.exports = class LevelView {
...
+          } else {
+            sprite.position.y = sprite.cameraOffset.y;
+          }
+        }, this);
+
+        // Play that thing!
         this.playAnimationWithOffset(sprite, "idle", Math.floor(Math.random() * 3) + 21, 1);
         break;
-
+      }
```
joshlory: Nit: braces around case statement aren't needed.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -2096,28 +2099,69 @@ module.exports = class LevelView {
...
+        frameList = frameList.concat(backBuffer);
+
+        let animation = sprite.animations.add("idle", frameList, 5, true);
+
+        // onLoop, we want to randomize which corner of the index the animation manifests in.
+        animation.onLoop.add((sprite) => {
+          let xType = Math.floor(Math.random() * 2);
+          let yType = Math.floor(Math.random() * 2);
+          if (xType) {
+            sprite.position.x = sprite.cameraOffset.x + 20;
```
joshlory: What is the `sprite.cameraOffset` value?  My concern is that this might be tied to the current camera location and move on larger 20x20 levels.  Can we calculate the the corner positions from the `x` and `y` values passed in to `createBlock`?

src/js/game/LevelMVC/LevelView.js
```diff
@@ -2096,28 +2099,69 @@ module.exports = class LevelView {
...
+            anim = Phaser.Animation.generateFrameNames("redstone_sparkle", 16, 23, ".png");
+            break;
         }
-        frameList = frameList.concat(Phaser.Animation.generateFrameNames("redstone_sparkle", 8, 13, ".png"));
-        for (i = 0; i < 3; ++i) {
-          frameList.push("redstone_sparkle13");
+
+        // Set up a random amount of blank frames to be put on the front and back for delay between loops.
+        let frontDelay = Math.floor(Math.random() * 50) + 5;
+        let backDelay = Math.floor(Math.random() * 30) + 20;
```
joshlory: Nit: let's pull these out as a `randomInt(min, max)` helper function.

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: I didn't think so either, but the linter won't let that fly, given that I declare local scoped variables like:
        let frontBuffer = [];
        let backBuffer = [];

you get the error:
2103:7   error  Unexpected lexical declaration in case block  no-case-declarations

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: Yeeeeeeah, I have that concern as well. What I'm using it for there is that when you go into that anonymous function as a part of the onLoop call, I was losing passed in arguments on subsequent calls. Basically, position changes, so that's not a reliable value to grab in terms of an offset, cameraOffset however was never changing.

Basically, cameraOffset was always the center of the index and I used it as an anchor point to math out where I want the sprite to be drawn (which is captured in position). On subsequent loops, position will have changed and you'd get the sprite just drifting away unless you can reference that original anchor point somehow?

I have the same concern about moving cameras though... but perhaps it'd be fine, because we'd want the sprite to ""move"" like the rest of the index to a different position on the screen? I think this ought to be tested with a big map.

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: caaaan do 👌

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: @joshlory confirmed that using the camera offset results in the correct behavior.

my guess is that it works something like:
>position means draw to screen with some math: 40 * x, 60 * y (size of each sprite)
>but that's wrong if the camera isn't focused on the same spot, on a huge map 40 * x index, if x were say = 5, shouldn't draw in the center of the player area if we're way off on some corner of the map, so take the camera's position into account.
>sparkle is now doing the same thing, taking the camera position into account.

Now that we're parenting them, I bet you could say something like getParent() -> using parent index in levelModel somehow to find its sprite -> go to actionGroup to find the appropriate parent sprite -> use THAT sprites position as an offset for the child sprite...

But in doing that I imagine you'd just get the same numbers: the position -> screen coords + camera offset.

I think it's fine to skip all that hubbub and just use the math we know is being used anyway? As I said, I mad a 10x13 map and walked around with the camera panning and the sparkle always plays in the right position in relation to the redstoneWire that's generating it. 

👌

src/js/game/LevelMVC/LevelView.js
joshlory: Ah, interesting.  That makes sense then!

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1025,6 +1025,10 @@ module.exports = class LevelView {
       const group = block.shouldRenderOnGroundPlane() ? this.groundGroup : this.actionGroup;
       const offset = block.shouldRenderOnGroundPlane() ? -0.5 : 0;
       sprite = this.createBlock(group, position[0], position[1] + offset, blockType);
+      if (blockType.startsWith('redstoneWire') && blockType.endsWith('On')) {
+        // Please excuse the magic numbers. They're just to offset the sparkle correctly within the index.
+        sprite.addChild(this.createBlock(group, 0.35, 0.05, "redstoneSparkle"));
+      }
```
joshlory: Actually, I think we want to move this logic to `createBlock`?  So that however createBlock is called it will always create the sparkle animation for powered wires.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -2247,6 +2248,12 @@ module.exports = class LevelView {
         break;
     }
 
+    // If we're making a charged redstoneWire, we need to do this again to make the sparkles
+    if (blockType.startsWith('redstoneWire') && blockType.endsWith('On')) {
+      // Please excuse the magic numbers. They're just to offset the sparkle correctly within the index.
+      sprite.addChild(this.createBlock(group, 0.35, 0.05, "redstoneSparkle"));
```
joshlory: I think this actually causes us to unnecessarily add the sprite to the plane group.  Here's what I was suggesting: https://github.com/code-dot-org/craft/pull/398 (I need to resolve conflicts for the most recent `randomInt` changes).

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1987,6 +1988,10 @@ module.exports = class LevelView {
       return sprite;
     };
 
+    if (blockType.startsWith("redstone")) {
+      console.log("whgat");
+    }
```
joshlory: Extra?

PR #385, opened 2017-10-26
player should now not be able to move before running

WilliamHarmonMS:
@joshlory 

This should be a quick and dirty way to ensure the player isn't moving before the run button is clicked. Also not sure how to visualize this change, since the dev environment doesn't have a run button, but it works to prevent his movement if run doesn't set to true (and manually calling controller.run() does turn the player "on").

it seems that controller.run() is just called on load up for the http://localhost:3017/ page though, so it might be a good idea to -check- that somewhere?


WilliamHarmonMS:
Totally agree that reverting the previous PR is a better way to go. 
@alalb may have more information about the justification for this change.

WilliamHarmonMS:
@Hamms or @joshlory 

(This one specifically should technically be trashed and simply have #248 reverted. Have just left it up so we don't lose track of the need for the fix)

joshlory:
Can you update this PR so it is a revert of https://github.com/code-dot-org/craft/commit/6d397631cd6cbebf4d9aa25d7731fa1342f4a2a3?  You'll want to pass the `-m 1` option to revert the merge.

WilliamHarmonMS:
Absolutely 👍 

I believe this branch is now in a good state? Pulled updated master branch and reverted #248 so I believe approval here would just be master - #248 without the original changes I made in 7c7ccb3

joshlory:
Change looks good from a code perspective, thanks!  I'd like to understand the motivation for this change before merge.

As it works now the Player is free to explore before clicking "Run", but on level 1 they will be trapped in the house so they can't go very far.  On the final free play level, letting the player move will let them explore off-screen areas so they can plan ahead what code they want to write for the Agent.

alalb:
@joshlory This came from Deirdre. The goal is to write code before they can do anything in the level, since we are adding code to level 1, we want to encourage students to code the agent to solve problems then navigate the player around.

joshlory:
Hmm, was this an issue you saw during in-classroom testing?

alalb:
Yes.

joshlory:
Hi Ali, I'm not sure I follow.  In level 1 the Player can't leave the house until the Agent opens the door.  What was the specific issue you observed in classrooms?  Thanks!

alalb:
@joshlory  This feature has been requested by Design and Deirdre and it has been decided that we want the player's movement to be locked before running the code. I don't think there are any blockers or issues preventing Devs from approving this one. 
Thanks!

joshlory:
> What was the specific issue you observed in classrooms? Thanks!

alalb:
Thanks @joshlory

joshlory:
Please note that a waived review is not an explicit approval.  Do you mind responding to the question above?

We can discuss the merits of both options, but I’d like it to be a collaborative decision.  Thanks!

alalb:
When level 1 was a movement intro level, we noticed that there is a steep learning curve on level 2 and students spent way more time that we wanted activating the pressure plate.

Deirdre: 
We identified 3 things contributing to the length of time spent on L2: 
-    Students continuing to use the mechanic we introduced on L1 which was just navigating with the Alex/Steve character
-    Level 2 becomes the first time they need to hit the run button
-    Role of the Agent isn’t clear

PR #386, opened 2017-10-26
use is back per Deirdre

WilliamHarmonMS:
Per Deirdre, use is back in. @joshlory 

@alalb  has more info


alalb:
Deirdre asked us to add the 'Use' functionality back to the game to shear sheep. with the infinite shearing issues fixed, I assume this will not cause any other issues.

PR #404, opened 2017-10-30
Bake AO offsets into the atlas JSON

joshlory:
Right now we offset sprites in the atlas JSON, then again in the code.  This PR combines these two offsets and bakes them into the atlas JSON, so no additional adjustment is needed in code.

In the process I found that the `AOeffect_Left` shadow was off by 1px to the east.  Before:

screen shot 2017-10-30 at 12 45 16 pm

Fixed:

screen shot 2017-10-30 at 12 44 49 pm

### DIFF
--- a/src/assets/images/LavaGlow.json
+++ b/src/assets/images/LavaGlow.json
@@ -5,7 +5,7 @@
 		"frame": {"x":0,"y":0,"w":12,"h":40},
 		"rotated": false,
 		"trimmed": true,
-		"spriteSourceSize": {"x":28,"y":22,"w":12,"h":40},
+		"spriteSourceSize": {"x":28,"y":0,"w":12,"h":40},
 		"sourceSize": {"w":64,"h":64},
 		"pivot": {"x":0.5,"y":0.5}
 	},
@@ -14,7 +14,7 @@
 		"frame": {"x":24,"y":16,"w":40,"h":8},
 		"rotated": false,
 		"trimmed": true,
-		"spriteSourceSize": {"x":0,"y":54,"w":40,"h":8},
+		"spriteSourceSize": {"x":0,"y":32,"w":40,"h":8},
 		"sourceSize": {"w":64,"h":64},
 		"pivot": {"x":0.5,"y":0.5}
 	},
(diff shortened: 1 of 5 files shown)


src/assets/images/LavaGlow.json
```diff
@@ -68,17 +68,17 @@
...
-		"spriteSourceSize": {"x":0,"y":22,"w":40,"h":16},
+		"spriteSourceSize": {"x":0,"y":0,"w":40,"h":16},
 		"sourceSize": {"w":64,"h":64},
 		"pivot": {"x":0.5,"y":0.5}
 	}},
 	"meta": {
 		"app": "http://www.codeandweb.com/texturepacker",
 		"version": "1.0",
-		"image": "WaterAO.png",
+		"image": "LavaGlow.png",
```
Hamms: why this change?

joshlory: Bad copy-paste by me earlier 😄.  The `LavaGlow.json` atlas is for `LavaGlow.png` (but it's the same layout as `WaterAO.png`, so I copied & pasted when doing the original alignment).

joshlory:
@WilliamHarmonMS can you ask Joel about this?

> I'm not super familiar with how the atlas JSON is generated; is there a chance this could get overwritten by some future update?

WilliamHarmonMS:
@joshlory 

My understanding of Joel's process is:
Make a bunch of files as individual .pngs -> program cobbles them together into a single file automatically placing them wherever it wants -> along with this file it generates the JSON file with specific pixel ranges that correlate to where each image was placed (in fact the JSON object names are generated using the names of the individually saved files)

Any manual changes to the JSON would thus not be included in future updates that Joel makes to those files -unless- there is a setting that he can include to mimic what you're trying to do. I'm not sure -how- robust his tools are, but I can definitely check up on that today. 👍

joshlory:
Thanks William!  The AO sprite sheet hasn't changed since it was added in 2015, so I think it's a safe bet that we can bake in the offsets.

If we do need to update the AO sprites in the future and Texture Packer can't import the new offsets from the AO.json file, I think it's ok to manually copy over the 5 `spriteSourceSize` adjustments.

Issue #406, opened 2017-10-30
Uncaught exception: "TypeError: Cannot read property 'onComplete' of undefined"


joshlory:
Repro:
 1. Navigate to https://code-dot-org.github.io/craft/.
 2. Place any block next to a powered redstone wire.

PR #413, opened 2017-10-30
Don't keep animating if the sprite is no longer in the scene

joshlory:
Fixes https://github.com/code-dot-org/craft/issues/406.


joshlory:
Weird that Phaser still calls `onComplete` if the sprite has been removed, but whatever ¯\\\_(ツ)\_/¯

PR #414, opened 2017-10-31
Fun with ghasts

WilliamHarmonMS:
@joshlory 

Cole and I had some fun. Deirdre was into it. What do you guys think? 😂 

(ghasts are meant to be placed at [1,5] and [7,8] as starting positions. This is a bit hard coded specifically to work with level 11 of 2017 HoC)

### DIFF
--- a/src/js/game/Entities/Ghast.js
+++ b/src/js/game/Entities/Ghast.js
@@ -1,19 +1,43 @@
 const BaseEntity = require("./BaseEntity.js");
 const randomInt = require("./../LevelMVC/Utils.js").randomInt;
 module.exports = class Ghast extends BaseEntity {
-    constructor(controller, type, identifier, x, y, facing) {
+    constructor(controller, type, identifier, x, y, facing, pattern = null, variant = null, primaryScalar = null, secondaryScalar = null, speed = null) {
         super(controller, type, identifier, x, y, facing);
         this.offset = [-50, -84];
         this.prepareSprite();
         this.sprite.sortOrder = this.controller.levelView.yToIndex(Number.MAX_SAFE_INTEGER);
         this.audioDelay = 15;
-        this.anchor = [this.sprite.position.x, this.sprite.position.y];
-        this.movingUp = false;
-        if (this.sprite.position.x > 60) {
-          this.movingUp = true;
+        if (pattern !== null) {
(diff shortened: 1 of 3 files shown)


WilliamHarmonMS:
funwithghasts

src/js/game/Entities/Ghast.js
```diff
@@ -87,26 +123,133 @@ module.exports = class Ghast extends BaseEntity {
...
+        break;
+      case "sinusoidal":
+        this.sinusoidal(this.details.variant);
+        break;
+      default:
+        //do nothing
+    }
+  }
+
+  sinusoidal(variant) {
```
joshlory: Can we use [`Phaser.Easing.Sinusoidal`](https://phaser.io/docs/2.3.0/Phaser.Easing.Sinusoidal.html) instead of rolling our own? 😄

src/js/game/Entities/Ghast.js
```diff
@@ -87,26 +123,133 @@ module.exports = class Ghast extends BaseEntity {
...
+    let destination = 0;
+    if (origin !== 3) {
+      destination = this.boxVertTracker + 1;
+    }
+
+    this.t += (this.speed / 60);
+    if (this.t > 1) {
+      this.t = 1;
     }
+    let pos = this.lerp(vertices[origin], vertices[destination], this.t);
```
joshlory: We should be leveraging Phaser tweens to move the sprite, rather than calculating the position ourselves.

joshlory:
Test coverage?

src/js/game/Entities/Ghast.js
WilliamHarmonMS: I did find it a bit absurd I had to write my own lerp function 😂 
Do phaser tweens have a lerp method that can be easily called? I can look into this for sure.

src/js/game/Entities/Ghast.js
WilliamHarmonMS: Does that do the same thing? My sinusoidal function is a bit tailored to the levelData format (The amplitude of the wave I don't suspect would be an issue, but the length of the path and the refersing behavior along the primary axis is something I don't -imagine- would be build in to a phaser easing function? Maybe though? I'll read the API

Issue #397, opened 2017-10-27
Rail does not have any cracking animation when being destroyed


v-chyou:
Found at: [https://levelbuilder-studio.code.org/s/craft17/stage/1/puzzle/10](https://levelbuilder-studio.code.org/s/craft17/stage/1/puzzle/10)

When the agent destroys a rail, cracks do not appear leading up to the cloud of dust when the rail finally breaks.  

1. Load map at address provided above
2. Program the agent: turn left, destroy block
3. Run code
4. Observe animation of rail being destroyed

Actual result: Cracks do not accumulate on rail in the time leading up to its destruction
Expected result: Cracks appear in rail as it is being destroyed

[railAnimation.zip](https://github.com/code-dot-org/craft/files/1423149/railAnimation.zip)

PR #418, opened 2017-10-31
phaser CAN crop

WilliamHarmonMS:
@joshlory or @Hamms 

Phaser can crop the destruction animation! Hurrah!

This crops the destruction overlay to the top 40x40 and then scoots it down 20 pixels so that it lines up over the index the player is destroying so that it can appropriately cover "flat" objects that need an overlay (basically only pressurePlates and rails).

this fixes [issue397](https://github.com/code-dot-org/craft/issues/397)

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -1156,9 +1156,9 @@ module.exports = class LevelView {
       // block immediately without waiting for the animation.
       afterDestroy();
     } else {
-      let destroyOverlay = this.actionGroup.create(-12 + 40 * destroyPosition[0], -22 + 40 * destroyPosition[1], "destroyOverlay", "destroy1");
+      const destroyOverlay = this.actionGroup.create(-12 + 40 * destroyPosition[0], -22 + 40 * destroyPosition[1], "destroyOverlay", "destroy1");
       if (LevelBlock.isFlat(blockType)) {
-        let cropRect = new Phaser.Rectangle(0, 0, 60, 40);
+        const cropRect = new Phaser.Rectangle(0, 0, 60, 40);
         destroyOverlay.position.y += 20;
         destroyOverlay.crop(cropRect);
       }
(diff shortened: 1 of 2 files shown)


src/js/game/LevelMVC/LevelView.js
```diff
@@ -1150,13 +1150,18 @@ module.exports = class LevelView {
...
+    if (LevelBlock.skipsDestructionOverlay(blockType)) {
       // "flat" blocks are by definition not cube shaped and so shouldn't accept
       // the cube-shaped destroy overlay animation. In this case, destroy the
       // block immediately without waiting for the animation.
       afterDestroy();
     } else {
-      const destroyOverlay = this.actionGroup.create(-12 + 40 * destroyPosition[0], -22 + 40 * destroyPosition[1], "destroyOverlay", "destroy1");
+      let destroyOverlay = this.actionGroup.create(-12 + 40 * destroyPosition[0], -22 + 40 * destroyPosition[1], "destroyOverlay", "destroy1");
+      if (LevelBlock.isFlat(blockType)) {
+        let cropRect = new Phaser.Rectangle(0, 0, 60, 40);
```
Hamms: nit: both of these `let`s can be `const`s

src/js/game/LevelMVC/LevelView.js
```diff
@@ -1150,13 +1150,18 @@ module.exports = class LevelView {
       this.playExplosionAnimation(playerPosition, facing, destroyPosition, blockType, completionHandler, true, entity);
     };
 
-    if (LevelBlock.isFlat(blockType)) {
+    if (LevelBlock.skipsDestructionOverlay(blockType)) {
```
Hamms: :+1:

src/js/game/LevelMVC/LevelView.js
WilliamHarmonMS: So they -can-
I was assuming that .crop() was modifying data somehow and that const would conflict with that, but that appears to -not be the case-.

Will update right now.

Hamms: `const` only limits your ability to assign a new value to the variable, not your ability to modify the already-assigned variable;

PR #425, opened 2017-11-02
integration tests now exist

WilliamHarmonMS:
All the integration tests have the optimal solution implemented for collecting the diamond and reaching the goal.

All levels have had data updated to reflect the new finished level design.

IMPORTANT: For some reason, despite following the exact same format, some tests time out. No idea why 🤷‍♂️


WilliamHarmonMS:
@joshlory and @Hamms 
this works. It all works.
Just don't look at the Timeout for Agent10
It works on WATCH=1 at 1000 ms, and on tap-diff at 7000 ms. I have NO IDEA why Travis can't pass it with anything less than this.

WilliamHarmonMS:
@joshlory @Hamms 

all tests currently passing. Going to add the other non-curriculum tests as another PR branched off this one, but this one looks green. I hope it's reliable?

PR #426, opened 2017-11-03
Allow Agent to place in front and behind, always on the Action Plane

WilliamHarmonMS:
@joshlory @Hamms 
So code freeze was when again? 😂


WilliamHarmonMS:
@joshlory 
PTAL, I think I covered all the bases for making placeDirection generic in all cases? I left placeForward / placeInFront in just because I wasn't sure what other years might depend on it (in terms of what previous year's blockly code is referencing).

let me know what you think about this PR. 👍

src/js/game/LevelMVC/FacingDirection.js
```diff
@@ -13,6 +13,14 @@ const FacingDirection = Object.freeze({
...
   },
 
+  left: function (facing) {
+    return (facing + 3) % 4;
+  },
+
+  right: function (facing) {
+    return (facing + 1) % 4;
+  },
+
```
Hamms: nit; there's a bit of overlap with these and the `turn(facing, rotation)` method; it'd be nice to update one of em to use the other.

src/js/game/GameController.js
```diff
@@ -1301,6 +1301,42 @@ class GameController {
     });
   }
 
+  placeBlockDirection(commandQueueItem, blockType, direction) {
```
joshlory: I think `placeBlockForward` is now just a special case of this method?  Can `placeBlockForward` call `placeBlockDirection` with the third param always the forward direction?

```js
placeBlockForward(commandQueueItem, blockType) {
  const entity = this.getEntity(commandQueueItem.target);
  const forwardPosition = this.levelModel.getMoveForwardPosition(entity);
  this.placeBlockDirection(commandQueueItem, blockType, forwardPosition);
}
```

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -345,8 +345,17 @@ module.exports = class LevelModel {
...
+    switch (direction) {
+      case FacingDirection.North:
+        return Position.forward(entity.position, entity.facing);
+      case FacingDirection.South:
+        return Position.forward(entity.position, FacingDirection.opposite(entity.facing));
+      case FacingDirection.East:
+        return Position.forward(entity.position, FacingDirection.right(entity.facing));
+      case FacingDirection.West:
+        return Position.forward(entity.position, FacingDirection.left(entity.facing));
+    }
```
Hamms: why is this assuming so strongly that the entity is always facing north, rather than just using `Position.forward(entity.position, direction)` for all cases?

src/js/game/LevelMVC/FacingDirection.js
```diff
@@ -13,6 +13,14 @@ const FacingDirection = Object.freeze({
     }
   },
 
+  left: function (facing) {
+    return (facing + 3) % 4;
```
joshlory: Nit: let's deduplicate this with `turn`.  Either implement as `return FacingDirection.turn(facing, 'left')` or update `turn` to:

```js
turn: function (facing, rotation) {
  return FacingDirection[rotation](facing);
}
```

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -583,21 +592,37 @@ module.exports = class LevelModel {
     return entity.canPlaceBlock(blockAtPosition);
   }
 
+  canPlaceBlockDirection(blockType = "", entity , direction) {
+    if (entity.isOnBlock) {
+      return false;
+    }
+    let plane = this.getPlaneToPlaceOn(this.getMoveDirectionPosition(entity, direction), entity);
+    if (plane === this.groundPlane) {
+      if (blockType === "redstoneWire" || blockType.substring(0,5) === "rails" && this.groundPlane.getBlockAt(this.getMoveDirectionPosition(entity, direction))) {
```
Hamms: this logic _definitely_ needs to live somewhere else

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -583,21 +592,37 @@ module.exports = class LevelModel {
...
+        return false;
+      }
+    }
+    return this.getPlaneToPlaceOn(this.getMoveDirectionPosition(entity, direction), entity) !== null;
+  }
+
   canPlaceBlockForward(blockType = "", entity = this.player) {
```
joshlory: Same feedback as above — I think this becomes a special case of `canPlaceBlockDirection` (i.e. this method should just call that implementation).

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -695,6 +720,17 @@ module.exports = class LevelModel {
     return targetPlane.setBlockAt(blockPosition, new LevelBlock(blockType));
   }
 
+  placeBlockDirection(blockType, targetPlane, entity, direction) {
```
joshlory: Same feedback for `placeBlockForward`.

src/js/game/GameController.js
WilliamHarmonMS: Yeah, that's totally fine. I wasn't messing with placeBlockForward simply because I didn't know what assumptions were going to be made with previous years and what not, but that's actually a great suggestion.

Will implement. 👍

src/js/game/LevelMVC/FacingDirection.js
WilliamHarmonMS: Do we want opposite to be lumped in there as well? They all feel very similar to me. "Given the direction I'm facing, give me another direction relative to that one.
>left of
>right of
>reverse of

Changing the left/right to call turn is easy but let me know if you feel reversing should be lumped in there as well.

src/js/game/LevelMVC/LevelModel.js
WilliamHarmonMS: It's not assuming the entity is facing north?

for args: entity is the entity doing the check (this is currently used in placeDirection only I believe), and direction is the direction relative to that entity (so 2 for instance would not be absolute down, it would indicate -behind- whatever direction the entity is facing).

So if direction === 2, our case hits FacingDirection.South, which is honestly the worst part of this, but I tried to use the FacingDirection enum there rather than just numbers so we would have some semblance of what they mean.

In case FacingDireciton.South:
we're going to return position.Forward (the index "in front" of the entity) given the entity position, and the facing direction relative to that entity. 

So for behind that's: FacingDirection.opposite(entity.facing)); meaning the index that is opposite the direction the entity is facing.

Or left, for case West:, right for case East:, and so on.

src/js/game/GameController.js
WilliamHarmonMS: @joshlory 
Also, placeBlockDirection() doesn't take a position for it's third arg, it takes a relative direction. So this is sufficient: 😂 
>>>  placeBlockForward(commandQueueItem, blockType) {
    this.placeBlockDirection(commandQueueItem, blockType, 0);
  }

src/js/game/LevelMVC/LevelModel.js
WilliamHarmonMS: Moved it into a static method in LevelBlock.

This cool? 👍

WilliamHarmonMS:
@joshlory @Hamms 

addressed the PR feedback. Only question I have is in regards to whether or not we like FacingDirection.getRelativeDirection()

it's certainly clear, but really niche? It's only used in a single case, but I feel kinda icky just comparing strings.

src/js/game/LevelMVC/FacingDirection.js
```diff
@@ -23,7 +40,7 @@ const FacingDirection = Object.freeze({
   },
 
   turn: function (facing, rotation) {
-    return (facing + 4 + (rotation === 'right' ? 1 : -1)) % 4;
+    return (facing + 4 + (rotation === 'right' ? 1 : -1)) % 4; // still speechless that JS has a mod bug...
```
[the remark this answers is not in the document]
WilliamHarmonMS: haha, didn't mean to leave that in but:
The ... + 4 .... % 4 baffled me, because that literally should do nothing (as 1 % 4 === 5 % 4)
BUT
it's to keep the a value (of a%b) in the positive range, because js returns incorrect mod values when a < 0.

-1 % 4 = 3
(because a%b = [a/b - floor(a/b)] * b
or in this case: -1 % 4 = [-1/4 - floor(-1/4)] * 4
                                       [-1/4 - (-1)] * 4
                                        [3/4] * 4
                                              3

yet in js -1 % 4 returns -1, and I would bet MONEY that it's because the operator has a 
if (a < b) {return a;} check, which is appropriate with positive numbers, but not negative numbers.

😂 

just chuckles.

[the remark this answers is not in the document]
Hamms: Oh, it's actually because `%` in JS is [the _remainder_ operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators#Remainder) not actually modulo

PR #429, opened 2017-11-07
Integration noncurriculum

WilliamHarmonMS:
@joshlory @Hamms 

Added "functionality tests" to cover piston entity conflict resolution and the rail riding.

NOTE: This is branched off the other integration test branch, so as not to -mandate- that the curriculum level tests be merged at the same time as these. 

Double NOTE: For some reason the Rails test is a problem again. I have no clue why and I'm starting to think it may be -more- than a timing thing. This wait -more- than enough time for it to complete were it just that. Perhaps something with the animation systems when run on here? I'm a bit out of ideas, so we should probably discuss these?


WilliamHarmonMS:
@joshlory @Hamms 
OK

So with that last commit, Rails functionality is actually passing.

Agent10 still does't pass but it's more than just waiting for the Agent to do stuff. Functionality02 (rails) took a long time to get going, and even at 15x speed had to have a delay of 10 seconds. My guess is something with this version isn't running the animations at 15x speed? Ever? The fact that is passes locally is weird, as they're both Headless Chrome 0.0.0, so unless (Mac OS X 10.12.6) vs Linux (0.0.0) makes a difference, I'm a bit stumped.

Point is, with that last functionality test -rails have a working integration test-

Even though we don't have a passing test exactly mimicking agent10, with this PR we will have all 2017 functionality covered with integration tests 👍 👍 👍

WilliamHarmonMS:
Also I can squash a bunch of these commits if we want... Having to push to see what Travis CI thought (since all local test runs passed the whole time) made it a bit messy. Let me know about that.

joshlory:
These look great!  I think squashing into a smaller set of commits sounds good.  You can also rebase on `master` after merging PR https://github.com/code-dot-org/craft/pull/425 to remove that part of the diff.

Worth removing the flowers from the ground decoration plane? 😄

screen shot 2017-11-07 at 9 50 03 am

WilliamHarmonMS:
This got too messy, so I just made a new branch that's much cleaner.

PR #430, opened 2017-11-07
Function tests

WilliamHarmonMS:
This is replacing PR #429 because that commit chain got so messy. This is a cleaned up branch.


WilliamHarmonMS:
@joshlory

test/helpers/AgentLevels.js
```diff
@@ -138,13 +138,91 @@ module.exports = {
...
+      "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "",
+      "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "",
+      "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "",
+      "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "",
+      "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "",
+      "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "",
+      "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "",
+      "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "",
+      "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "",
+      "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", ""],
```
joshlory: Should the 20x20 level have data from the real Level 12?

test/integration/AgentTest.js
```diff
@@ -508,7 +508,7 @@ test('Agent 9: Clear Path', t => {
   }));
 });
 
-/*test.only('Agent 10: Ride Rails', t => {
+/*test('Agent 10: Ride Rails', t => {
```
joshlory: Changes to this commented-out test? 😛

WilliamHarmonMS: Well I didn't want to forget it when someone removes the comments later, provided that test -ever- works 😭

test/helpers/AgentLevels.js
WilliamHarmonMS: This was 100% me being lazy, BUT I don't think we actually lose anything. The idea behind this integration test was to demonstrate that a 20x20 level is traversable, which it does. (So, for instance, position [15,15] is valid here, where it wouldn't be in a 10x10).

The -reason- I was lazy was because Craft.gameController.levelData, what I had been doing to grab the plane data on the other levels, truncates the list of strings when they're over 100. like so:

"bricks", "", "", "", "", "", "", "", "", "", "", "", …]

So rather than be able to rip the whole plane via copy/paste, I'd have to go through and manually populate each index... which I super didn't feel like doing.

Given that this test does ensure that the player can traverse a whole 20x20 grid, I think it -is- achieving what we want? Unless we had other ideas for what it should be doing? Block placing? Destroying? All that functionality is tested elsewhere, so maybe I'm just not 100% clear on what this test should be testing?

PR #431, opened 2017-11-09
handle plate if getting on rails

WilliamHarmonMS:
moved rail riding handling to entity class, and included a pressureplate handling


WilliamHarmonMS:
@joshlory 
this one just seemed really easy. PTAL ❤️

Issue #424, opened 2017-11-02
Riding the minecart on a track as it is being destroyed causes a crash


v-chyou:
Found at: [https://levelbuilder-studio.code.org/s/craft17/stage/1/puzzle/10](https://levelbuilder-studio.code.org/s/craft17/stage/1/puzzle/10)

If the player is riding a minecart on to a section of track that is being destroyed by the agent, the game crashes.

1. Load Level 10
2. Program the agent to destroy and place rails using the functions provided
3. Run the code
4. Ride the minecart while the code is still running
5. Ride on to a section of rail that is being destroyed by the agent

Actual result: Crash
Expected result: No crash
[railCrash.zip](https://github.com/code-dot-org/craft/files/1439159/railCrash.zip)

code-studio-common.min-1ba6ba760ec6d7130769edbf68380bb2a488e1e35fa85c25f1271c1eb2a283f1.js:4 The key "target-densitydpi" is not supported.
it.fixViewportForSmallScreens_ @ code-studio-common.min-1ba6ba760ec6d7130769edbf68380bb2a488e1e35fa85c25f1271c1eb2a283f1.js:4
craft.min-c38f8614ec0d60ef670d58ae82a0455e80da34b43f74aefd29da9b1fc5101b83.js:12 Uncaught TypeError: Invalid attempt to destructure non-iterable instance
    at craft.min-c38f8614ec0d60ef670d58ae82a0455e80da34b43f74aefd29da9b1fc5101b83.js:12
    at e.exports.e.playTrack (craft.min-c38f8614ec0d60ef670d58ae82a0455e80da34b43f74aefd29da9b1fc5101b83.js:13)
    at craft.min-c38f8614ec0d60ef670d58ae82a0455e80da34b43f74aefd29da9b1fc5101b83.js:13
    at craft.min-c38f8614ec0d60ef670d58ae82a0455e80da34b43f74aefd29da9b1fc5101b83.js:14
    at c.SignalBinding.execute (phaser.min-86cb81ba30b6548a6f477d373370222ff3e168bc38caf389bc065ec894437fc8.js:sourcemap:9)
    at c.Signal.dispatch (phaser.min-86cb81ba30b6548a6f477d373370222ff3e168bc38caf389bc065ec894437fc8.js:sourcemap:9)
    at c.Tween.update (phaser.min-86cb81ba30b6548a6f477d373370222ff3e168bc38caf389bc065ec894437fc8.js:sourcemap:15)
[...]

PR #432, opened 2017-11-09
this might fix it

WilliamHarmonMS:
All I can think of for [issue424](https://github.com/code-dot-org/craft/issues/424)

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -736,6 +736,12 @@ module.exports = class LevelView {
     // and we need to grab that track -in front of us-
     if (track === undefined) {
       track = this.controller.levelModel.actionPlane.getMinecartTrack(nextPos, facing);
+      // Having a weird bug on publish where rail destruction while riding causes a destructure of
+      // non-iterable instance error. If getTrack fails with currPos and nextPos, just call the whole thing off.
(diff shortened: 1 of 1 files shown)


WilliamHarmonMS:
@joshlory @Hamms 
This is a shot in the dark based on my comment on the issue page. But.... who knows? Apparently the bug is somewhat reliably repo-able on the published site, but my dev environment is not exhibiting any odd behavior.... 🤔

Hamms:
Does riding on rails that were placed at level initialization vs rails placed by agent code make a difference?

WilliamHarmonMS:
that is.... an excellent question. I think in the actual bug video it's a rail that was there at init? I've still no idea how that could make a difference. My best guess is that it's something like
Is there a forward Pos rail? Yes. -> but it's being destroyed -> so I arrive there, and assume there will then be a current Pos rail -> but there isn't anymore so we crash.

The line I linked in the issue could have problems there because we would check nextPos and then if that was undefined or null, handle it, but then we checked currPos later and assumed it would work. The fix I put in no longer makes that assumption and says: 
"If nextPos failed, get currPos because this might be out first rail. If currPos also fails (so next and curr are both undefined) we shouldn't even be in here, so abort the whole process.

I -think- that should work? But again, it's a really weird scenario that's hard for me to test. 😞

Issue #402, opened 2017-10-30
Small visual artifact on some miniblocks


joshlory:
![image](https://user-images.githubusercontent.com/413693/32186772-4b1408dc-bd60-11e7-895b-a04f1ca1803f.png)

https://levelbuilder-studio.code.org/s/craft17/stage/1/puzzle/8

WilliamHarmonMS:
fix was merged so closing this unless it reappears.

PR #433, opened 2017-11-09
no more shadow artifacts

WilliamHarmonMS:
@joshlory @Hamms 
This is the latest sprite sheet that Joel got me, and it should be fixing the small artifacts that sometimes show up, as described in [issue402](https://github.com/code-dot-org/craft/issues/402)


Hamms:
This will likely require the same kind of special cache flush when we deploy as do significant changes to `Blocks.png`

PR #442, opened 2017-11-14
new ghasts from joel! hopefully no more artifacting?

WilliamHarmonMS:
@joshlory 

Could you check this branch out and make sure it works with the new tween driven flight functionality? I'm not sure how that's done with the new phaser calls rather than my math. 🙇


Hamms:
@joshlory this is also going to require a cache flush when we deploy

PR #443, opened 2017-11-14
weird door bug where the animation and the walkability of a door coul…

WilliamHarmonMS:
…d become desynched

@joshlory 

This is a weird one, but it pertains to this Trello item:
https://trello.com/c/sNi5uJWW/130-level-2-bug-with-door-to-solve-the-puzzle-cant-walk-through-even-though-it-is-open

This change just asks if the door should be open and sets the walkability to true if that's the case (just in case it wasn't true before).


WilliamHarmonMS:
Also wut am I looking at with this Travis-CI failure? Something with Ghast.js? This PR has nothing to do with Ghast.js.... what's up with that Travis?

PR #444, opened 2017-11-15
door animation walkability could be desynched

WilliamHarmonMS:
Fixing the trello issue:
https://trello.com/c/sNi5uJWW/130-level-2-bug-with-door-to-solve-the-puzzle-cant-walk-through-even-though-it-is-open

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -2280,7 +2280,7 @@ module.exports = class LevelView {
     const position = this.controller.levelModel.actionPlane.indexToCoordinates(index);
     this.playDoorAnimation(position, open, () => {
       const block = this.controller.levelModel.actionPlane.getBlockAt(position);
-      block.isWalkable = !block.isWalkable;
+      block.isWalkable = block.isOpen;
       if (block.blockType !== "doorIron") {
         // Iron doors don't need to set the player animation to Idle, because they're not opened with 'use'.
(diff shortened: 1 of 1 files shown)


WilliamHarmonMS:
We probably could. The check I put in was just a catch-all for if the animation state was ever desynched from the walkability state. Putting it where you referenced would be trying to make sure they never get desynched in the first place, which I can definitely do. 👍

Hamms:
Ah, that makes sense. Looks like we're managing walkability as a strict toggle:

https://github.com/code-dot-org/craft/blob/c77946f08aa24d5ef95c199af3865f6098cb162d/src/js/game/GameController.js#L927

Does replacing that line with `frontBlock.isWalkable = canOpen` fix the desync?

WilliamHarmonMS:
I 100% bet it will. The original bug is actually pretty tough to repro, but when you did manage to get it, throwing a breakpoint in would show that only walkability was ever incorrect. isOpen was always indicative of the animation state (which is exactly what it should be). So that toggle there should 100% fix it. I'll do some quick testing. 👍

src/js/game/LevelMVC/LevelModel.js
```diff
@@ -532,6 +532,9 @@ module.exports = class LevelModel {
     var result = [false,];
 
     if (this.inBounds(position)) {
+      if (this.actionPlane.getBlockAt(position).isOpen) {
+        this.actionPlane.getBlockAt(position).isWalkable = true;
+      }
```
Hamms: can this change be safely removed now?

Issue #446, opened 2017-11-15
Sand does not fall when placed over water


v-chyou:
Found at: [https://studio.code.org/s/hero/stage/1/puzzle/12](https://studio.code.org/s/hero/stage/1/puzzle/12)

If the agent uses a place block with a direction (not down) to place sand over water, the sand block does not drop into the water.  This is not at parity with Minecraft.

Repro Steps:
1. Load the free-play level with new omni-directional place blocks at the address above
2. Program the agent to place a sand block over the water
3. Observe sand block

Actual result: Sand block does not fall into water
Expected result: Sand block falls into water
[sandParity.zip](https://github.com/code-dot-org/craft/files/1475665/sandParity.zip)

WilliamHarmonMS:
[PR449](https://github.com/code-dot-org/craft/pull/449) was closed so this is fixed.

Issue #445, opened 2017-11-15
Place direction block does not take into account blocks that are already in those locations.


v-jaarba:
Found in: https://studio.code.org/s/hero/stage/1/puzzle/12

This does not follow the same rules set that the other place block follows. If a block is in the current location it should have to be destroyed before anything else can be placed in that location. This will not work in Minecraft and will lead to imported code not working users.

Repro Steps:
1) Start HoC and proceed to level 12.
2) Make code to place two different blocks forward.
3) Run code and Observe.

Expected Results:
If a block is currently in the location the place direction block will not place anything in that location.

Actual Results:
Place Direction blocks will replace anything in a location with a new block.

Video:
[Microsoft Edge 11_15_2017 9_00_40 AM.zip](https://github.com/code-dot-org/craft/files/1475649/Microsoft.Edge.11_15_2017.9_00_40.AM.zip)

WilliamHarmonMS:
[PR449](https://github.com/code-dot-org/craft/pull/449) was closed so this is fixed.

Issue #447, opened 2017-11-15
Agent's placement animation does not complete when using an omni-directional place block


v-chyou:
Found at: [https://studio.code.org/s/hero/stage/1/puzzle/12](https://studio.code.org/s/hero/stage/1/puzzle/12)

When the agent places a block with any direction in the drop down menu, its right arm does not completely return to its side.  If the agent performs any other actions, this problem is corrected.  But if the agent's last action is to place a block in a non-down direction, the incomplete animation is quite apparent.

Repro Steps:
1. Load free-play level with omni-directional place blocks at address above
2. Program the agent to place a block in any direction other than down
3. Observe agent's right arm at end of animation

Actual result: Animation does not finish
Expected result: Animation finishes appropriately.
[armAnimation.zip](https://github.com/code-dot-org/craft/files/1475735/armAnimation.zip)

WilliamHarmonMS:
[PR449](https://github.com/code-dot-org/craft/pull/449) was closed so this is fixed.

PR #449, opened 2017-11-15
Agent placeDirection bugged behavior

WilliamHarmonMS:
This fixes the bugs with Agent placeDirection:
(1) Agent placeDirection was replacing other blocks in the Action Plane [issue445](https://github.com/code-dot-org/craft/issues/445)
(2) Agent placeDirection was placing sand/gravel in the wrong plane when over water [issue446](https://github.com/code-dot-org/craft/issues/446)

These should both be fixed with this PR.

@joshlory

### DIFF
--- a/src/js/game/GameController.js
+++ b/src/js/game/GameController.js
@@ -1177,7 +1177,7 @@ class GameController {
     }
 
     position = this.levelModel.getMoveDirectionPosition(player, direction);
-    placementPlane = this.levelModel.getPlaneToPlaceOn(position, player);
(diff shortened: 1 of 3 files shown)


WilliamHarmonMS:
Also just fixed [issue447](https://github.com/code-dot-org/craft/issues/447)

I think I introduced that one when I refactored placeDirection as a whole. Whoops.

WilliamHarmonMS:
@Hamms PTAL if you can 👍

WilliamHarmonMS:
Last commit fixes [issue450](https://github.com/code-dot-org/craft/issues/450)

PR #452, opened 2017-11-20
bounds checking crash on rail riding

WilliamHarmonMS:
@joshlory  or @Hamms 

Apparently there is a crash when the user:
(1) is riding a rail that leads them right up against the edge of the map (so that the "next index" would be out of bounds)
AND
(2) you destroy the final rail segment after they've begun moving to it (cleared the "next rail" being valid check) but the segment is destroyed by the Agent before the player reaches that index

Whackiness but... it's definitely a bounds check on nextPos there. Figured I'd throw in a bounds check for currPos as well because who knows? Should be impossible but that'll be extra safe.


WilliamHarmonMS:
Yeah no worries, I can make that change and then merge.  👍

PR #461, opened 2018-01-16
update Tests to use new Position instances

Hamms:


joshlory:
Will this require us to update how .level file positions get translated to config options?  For example:  https://github.com/code-dot-org/code-dot-org/blob/2e05802f4942b61873ee1edbeaf7260fa5c6df29/dashboard/config/scripts/levels/Overworld%20Free%20Play%2020x20.level#L12

Hamms:
Nope! I make sure to [use `Position.fromArray`](https://github.com/code-dot-org/craft/pull/461/files#diff-82ea2716b14e187d2ef963445c025e64R67) for all of those.

Note that the tests [are using the same config](https://github.com/code-dot-org/craft/blob/master/test/helpers/AdventurerLevels.js#L7), so we'd know if we needed to update

Hamms:
I also use Position.fromArray for those methods that are used by the per-level success and failure methods: https://github.com/code-dot-org/craft/pull/461/commits/d61c99e10ebb9763567d38c822b4542d9a127034

PR #467, opened 2018-06-08
Build with webpack

Hamms:


src/index.html.ejs
```diff
@@ -1,176 +0,0 @@
...
-  document.write(
-    '<script src="//'
-    + window.location.hostname
-    + ':<%= liveReloadPort %>/livereload.js"><\/script>'
-  );
-</script>
-<% } %>
-
-</body>
-</html>
```
joshlory: Nit: this cleanup could be pre-flighted in a separate PR to reduce the diff.

webpack.config.js
```diff
@@ -0,0 +1,33 @@
...
+    filename: '[name].js',
+    libraryTarget: 'umd',
+  },
+  module: {
+    rules: [{
+      test: /\.js$/,
+      loader: "babel-loader",
+    }, {
+      test: /\.(png|svg|jpg|gif|json|mp3|ogg|txt|wav)$/,
+      loader: 'file-loader'
```
joshlory: Tiny nit: missing trailing comma (we're inconsistent within this file).

joshlory:
Worth adding a [`performance.assetFilter`](https://webpack.js.org/configuration/performance/#performance-assetfilter) to reduce some of the size limit warnings?

src/index.html.ejs
Hamms: I'm actually not sure it could; we're able to move this file from `index.html.ejs` to `index.html` because we are able to get rid of the `liveReloadPort` stuff here at the end, but we're only able to do that because webpack dev server provides it for us automatically

Hamms:
PTAL!

src/index.html.ejs
joshlory: Ah, my bad!

webpack.config.js
```diff
@@ -0,0 +1,38 @@
+const CopyWebpackPlugin = require('copy-webpack-plugin');
+const path = require('path');
+
+module.exports = {
+  entry: {
+    main: './src/index.js',
+    levels: './src/levels.js'
```
joshlory: Tiny style nit: missing trailing comma.

PR #481, opened 2018-09-27
Treasure chests

joshlory:
Chests open when a player stands next to them (orthogonally).


joshlory:
I should pay attention to the Codecov Report.  Accidentally merged with `test.only`, which only runs one test in the given test file.

https://github.com/code-dot-org/craft/blob/c60f95ec33e9a30f270ef0b63ad45cf3276db12c/test/unit/EntityTest.js#L93

Fixed in commit https://github.com/code-dot-org/craft/commit/1435e388baa1b87052582451ff5e4e610d907cf1.

PR #484, opened 2018-09-28
Lock to webpack 4.19

joshlory:
Thank you @madelynkasula for finding & tracking down a fix!


madelynkasula:
thanks, josh! referencing the [webpack issue](https://github.com/webpack/webpack/issues/8082) for future use, too

PR #486, opened 2018-10-01
Smooth prismarine block alpha transition

### LINKED ISSUE #480: Newly-placed prismarine blocks don't change color in sync with existing prismarine
See aquatic level 11.

Also, prismarine blocks should fade evenly between colors, instead of switching stepwise.

madelynkasula:
Fixes #480.

We wanted the prismarine block tweens to subtly change color, so this uses a yoyo'd `.to` transition rather than changing step-wise.

### Example
(forgive the dimensions on this gif...)
prismarine

I set the transition alpha to 0.5, but let me know if that's too drastic of a change!

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -2309,9 +2309,10 @@ module.exports = class LevelView {
         xOffset = this.blocks[blockType][2];
         yOffset = this.blocks[blockType][3];
         sprite = group.create(xOffset + 40 * x, yOffset + group.yOffset + 40 * y, atlas, frame);
-        frameList = Phaser.Animation.generateFrameNames("Prismarine", 0, 5, "", 0);
-        sprite.animations.add("idle", frameList, 5, true);
-        this.playScaledSpeed(sprite.animations, "idle", 0.1);
+        const options = [1500, Phaser.Easing.Sinusoidal.InOut, true, 0, -1, true];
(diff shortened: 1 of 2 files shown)


src/js/game/LevelMVC/LevelView.js
```diff
@@ -2309,9 +2309,10 @@ module.exports = class LevelView {
         xOffset = this.blocks[blockType][2];
         yOffset = this.blocks[blockType][3];
         sprite = group.create(xOffset + 40 * x, yOffset + group.yOffset + 40 * y, atlas, frame);
-        frameList = Phaser.Animation.generateFrameNames("Prismarine", 0, 5, "", 0);
-        sprite.animations.add("idle", frameList, 5, true);
-        this.playScaledSpeed(sprite.animations, "idle", 0.1);
+        const options = [1500, Phaser.Easing.Sinusoidal.InOut, true, 0, -1, true];
+        const tween = this.addResettableTween(sprite).to({
+          alpha: 0.5,
+        }, ...options);
```
Hamms: Is there a specific reason not to just write this as

```js
const tween = this.addResettableTween(sprite).to({
  alpha: 0.5,
}, 1500, Phaser.Easing.Sinusoidal.InOut, true, 0, -1, true);
```

madelynkasula: no, not specifically. i saw the convention elsewhere in the repo and thought it would be nice to clean up the options params a bit

PR #492, opened 2018-10-10
Player does not drown when stepping from land onto boat

### LINKED ISSUE #490: Player should not drown when stepping onto boat
Repro: Player steps from land onto boat
Expected: player stands in boat
Observed: player falls in the water, drowning animation plays

islemaster:
Fixes #490.

| Before | After |
| --- | --- |
| before | after |

### DIFF
--- a/src/js/game/LevelMVC/LevelModel.js
+++ b/src/js/game/LevelMVC/LevelModel.js
@@ -581,17 +581,13 @@ module.exports = class LevelModel {
       // Prevent walking into water/lava in levels where the player is
       // controlled by arrow keys. In levels where the player is controlled by
       // blocks, let them drown.
-      if (this.groundPlane.getBlockAt(position).blockType === "water") {
+      const blockTypeAtPosition = this.groundPlane.getBlockAt(position).blockType;
+      const frontEntity = this.getEntityAt(position);
+      if (['water', 'lava'].includes(blockTypeAtPosition)) {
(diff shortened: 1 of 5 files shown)


src/js/game/LevelMVC/LevelModel.js
```diff
@@ -581,17 +581,13 @@ module.exports = class LevelModel {
...
+          result.push(blockTypeAtPosition);
         } else {
-          return [true];
-        }
-      } else if (this.groundPlane.getBlockAt(position).blockType === "lava") {
-        if (this.controller.getIsDirectPlayerControl()) {
-          result.push("lava");
-        } else {
-          return [true];
+          return [frontEntity === undefined || frontEntity.canMoveThrough()];
```
Hamms: are we sure we want to end early here, and not with a `notEmpty` or `notWalkable` or something? It's not 100% clear to me that a value of just `[false]` is valid

islemaster: That's a good point - it'd be a smaller change if we _don't_ early-out when the new condition is not met.  I'll give that a shot.

islemaster:
:cry: Your change makes sense to me, but it broke a test.  Would you prefer we removed the last commit and got this fix in, or do you want me to go further right now trying to clean up this logic?

joshlory:
Ah I think that test may just be flakey at the 0.1 timescale.  It passed on re-run!

Issue #198, opened 2017-10-04
Line coverage can vary slightly even with no code changes


joshlory:
PR https://github.com/code-dot-org/craft/pull/52 attempted to reduce variance in code coverage metrics due to random animation timing.  This helped, but we're still seeing tiny variation in line coverage of animations in integration tests.

Creating this issue to track.

joshlory:
Sample diff: https://codecov.io/gh/code-dot-org/craft/pull/174/src/src/js/game/Entities/Chicken.js

joshlory:
Another: https://codecov.io/gh/code-dot-org/craft/pull/215/src/src/js/game/Entities/IronGolem.js

joshlory:
https://codecov.io/gh/code-dot-org/craft/pull/243/src/src/js/game/Entities/IronGolem.js

PR #499, opened 2018-10-11
Move all integration tests to new async pattern

### LINKED ISSUE #494: Use async / await for cleaner test code
You've already got it enabled! so we can write:

```js
test.only('Aquatic 2: Move to boat (pass)', t => {
  attempt('aquatic02', async (api, levelModel) => {
    // test
    t.end();
  }, 1);
});
```

instead of 

```js
test.only('Aquatic 2: Move to boat (pass)', t => {
  attempt('aquatic02', (api, levelModel) => new Promise(resolve => {
    // test
    t.end();
    resolve();
  }), 1);
});
```

islemaster:
Fixes #494. Follow-on to work in https://github.com/code-dot-org/craft/pull/495 and https://github.com/code-dot-org/craft/pull/498.

Applies new async/await pattern to all integration tests, completely removing the internal dependency on the `onAttemptComplete` callback argument to `startAttempt`.  However, I'm not entirely removing this argument yet - it's part of a public API and is still used in the main `code-dot-org/code-dot-org` repo.

### DIFF
--- a/demo/demo.js
+++ b/demo/demo.js
@@ -32,7 +32,7 @@ var gameController = new GameController({
   earlyLoadAssetPacks: testLevelToLoad.earlyLoadAssetPacks,
   earlyLoadNiceToHaveAssetPacks: testLevelToLoad.earlyLoadNiceToHaveAssetPacks,
   afterAssetsLoaded: () => {
-    gameController.codeOrgAPI.startAttempt(() => {});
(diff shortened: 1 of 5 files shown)


islemaster:
I don't quite understand how this PR affects coverage of the `IronGolem.js` file.

joshlory:
@islemaster I think it's related to issue https://github.com/code-dot-org/craft/issues/198.

PR #500, opened 2018-10-13
Use the correct animation when destroying blocks underwater

joshlory:


joshlory:
Bug: when destroying prismarine, the player gets caught up in `playItemAcquireAnimation` with `sprite.alive` set to false.  The first collection triggers the wrong completion handler (need to track down why).

joshlory:
image

PR #504, opened 2018-10-15
Shadowless miniblocks

joshlory:
Import new smaller miniblock spritesheet without shadow baked in.  Animate the spawn motion with a tween.

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -1454,21 +1454,19 @@ module.exports = class LevelView {
   }
 
   playItemDropAnimation(destroyPosition, blockType, completionHandler) {
-    var sprite = this.createMiniBlock(destroyPosition[0], destroyPosition[1], blockType);
+    let autoAcquire;
+    if (this.controller.getIsDirectPlayerControl() && completionHandler) {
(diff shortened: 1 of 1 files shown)


src/js/game/LevelMVC/LevelView.js
```diff
@@ -1523,21 +1454,19 @@ module.exports = class LevelView {
...
-    if (sprite) {
-      sprite.sortOrder = this.yToIndex(destroyPosition[1]) + 2;
-    }
-
-    if (this.controller.getIsDirectPlayerControl()) {
-      if (completionHandler) {
-        completionHandler();
-      }
+    let autoAcquire;
+    if (this.controller.getIsDirectPlayerControl() && completionHandler) {
```
islemaster: Can't tell if this is a logic error or an intentional change:
```js
// Before
if (this.controller.getIsDirectPlayerControl()) {
  completionHandler && completionHandler();
} else {
  // Happens when not direct player control
}

// After
if (this.controller.getIsDirectPlayerControl() && completionHandler) {
  completionHandler();
} else {
  // Happens when not direct player control OR no completion handler
}
```
Did you mean to change the conditions of the else case?

PR #512, opened 2018-10-17
Minecraft | treasure chests now drop items

WilliamHarmonMS:


src/js/game/LevelMVC/LevelView.js
```diff
@@ -430,6 +432,13 @@ module.exports = class LevelView {
       "conduit": ["blocks", "Conduit00", -12, 0],
 
       "chest": ["blocks", "Chest0", -12, -20],
+
+      "chestboat": ["blocks", "Chest0", -12, -20],
+      "chestnautilus": ["blocks", "Chest0", -12, -20],
+      "chestprismarine": ["blocks", "Chest0", -12, -20],
+      "chestheartofthesea": ["blocks", "Chest0", -12, -20],
```
joshlory: Rather than having a separate block for each chest type, can we follow the pattern we use for miniblocks?  The code converts `diamondMiniblock` to a diamond miniblock — I'd like to see `diamondChest` convert to a chest that drops a diamond miniblock.

WilliamHarmonMS: Aaah right I'd forgotten about the `diamondMiniblock`. Will that system be fine with the fact that we're just opening it rather than destroying the chest? I had a few different approaches I considered yesterday in terms of getting chests and different levels to dispense different treasure. 🤔 

I'll investigate the diamondMiniblock this morning.

WilliamHarmonMS:
Ugh 😩 

So you're right that's a way better method, but I was torn between "chest" and "Chest".

`diamondMiniblock` makes me want `diamondChest`, but since I'm pulling off the last 5 characters for the block to be spawned itself, that would means `"chest": ["blocks", "Chest0", -12, -20],` goes to `"Chest": ["blocks", "Chest0", -12, -20],` which breaks the block convention! There's no winning!

Actually a nit problem, and this is a way better method. Good call, my dude!

src/js/game/LevelMVC/LevelView.js
```diff
@@ -573,7 +576,18 @@ module.exports = class LevelView {
   playOpenChestAnimation(position) {
     const blockIndex = (this.yToIndex(position[1])) + position[0];
     const block = this.actionPlaneBlocks[blockIndex];
-    this.playScaledSpeed(block.animations, "open");
+    const animation = this.playScaledSpeed(block.animations, "open");
+    this.onAnimationEnd(animation, () => {
+      let blockData = this.controller.levelModel.actionPlane.getBlockAt(position);
+      let treasure = this.controller.levelModel.actionPlane.getBlockAt(position).blockType.substring(0, blockData.blockType.length - 5);
```
islemaster: For readability it'd be nice to extract a helper to hide away the substring logic here, so it reads something like
```js
const treasure = treasureTypeFromBlock(blockData);
```

src/js/game/LevelMVC/LevelView.js
```diff
@@ -429,7 +431,8 @@ module.exports = class LevelView {
       "bubbleColumn": ["blocks", "Bubble_Column0", -12, 0],
       "conduit": ["blocks", "Conduit00", -12, 0],
 
-      "chest": ["blocks", "Chest0", -12, -20],
+      "Chest": ["blocks", "Chest0", -12, -20],
```
joshlory: Note this is a breaking change; Cole will need to update puzzles that use "chest" once this ships.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -573,7 +576,21 @@ module.exports = class LevelView {
...
     const block = this.actionPlaneBlocks[blockIndex];
-    this.playScaledSpeed(block.animations, "open");
+    const animation = this.playScaledSpeed(block.animations, "open");
+    this.onAnimationEnd(animation, () => {
+      const treasure = this.getTreasureTypeFromChest(this.controller.levelModel.actionPlane.getBlockAt(position));
+      if (treasure) {
+        this.createMiniBlock(position[0], position[1], treasure, {
+          collectibleDistance: 1,
+          xOffsetRange: 10,
+          yOffsetRange: 10
```
joshlory: Let's override these both to 0, so the item always spawns centered on the open chest.

src/js/game/LevelMVC/LevelView.js
```diff
@@ -2079,6 +2110,12 @@ module.exports = class LevelView {
       };
 
       const collectiblePosition = this.controller.levelModel.spritePositionToIndex(offset, new Position(sprite.x, sprite.y));
+
+      // Chests will put out a treasure, but not disappear, so you need to be able to pick up set item from further away than the normal 1 index
+      if (blockType === "heartofthesea" || blockType === "boat" || blockType === "prismarine" || blockType === "nautilus") {
+        collectibleDistance *= 2;
+      }
```
joshlory: This is already overridable via the `overrides` argument.  But I think it's actually not bad to have the item show up and not collect it — then you can actually see what it is, vs. it disappearing right away.

src/js/game/LevelMVC/LevelView.js

I mean they should all get changed anyway to be <treasure>Chest right? No level has an empty chest? What a let down the students would find that! haha

Issue #334, opened 2017-10-18
Agent should "bob" over water instead of normal walking


joshlory:
Float up and down, with a small shadow below?  That's how miniblocks are drawn.

joshlory:
Guidance from Joel: start with no shadow, idle animation, +/- 3px offset for bob animation.

ryansloan:
won't fix

joshlory:
Re-opening for Aquatic 2018 — we want this same behavior for when the player is in a boat on the water.

joshlory:
+cc @madelynkasula

islemaster:
I've got this working for the boat entity (which doesn't move), but still trying to solve it for the player-in-a-boat.
![bob_after](https://user-images.githubusercontent.com/1615761/47103015-1e469380-d1f3-11e8-8813-e203d12044fd.gif)

PR #513, opened 2018-10-18
Boat bobs up and down

islemaster:
Fixes (at least part of) #334.

The boat bobs up and down on the water (3px easeInOutQuad running forward and back with a 2-second cycle).

This is actually two changes: One for the `Boat` entity, and one for the `Player` entity when its animations include the boat.

| `Boat` entity | `Player` in the boat |
| --- | --- |
|bob_after | peek 2018-10-17 16-46|

The latter required some refactoring!  Applying a tween to the Player sprite was problematic because (a) its position is tracked in world-space so any tweens affecting its position also had to be defined in world-space, and (b) we apply tweens to the player sprite position already to handle movement.

@joshlory suggested this might be solved with sprite parenting, and I went with this solution.  So now we've got an "animation rig" descendant of the player sprite which owns all of its animations, allowing us to apply a "bob" tween to the animation rig in player-local coordinate space without messing with any of the world-space tweens used to move the player around.

- World
  - `entity.sprite` whose position is defined in world-space (now contains no animations)
    - `entity.animationRig` whose position is relative to `entity.sprite` (contains all player animations)

This does mean a bunch of code in `LevelView` that was reaching into the entity's sprite's animation manager directly needs to be updated.  To avoid adding complexity all over the place, I'm having `BaseEntity` act as gatekeeper to its own animation manager.  This way any entity could be given an `animationRig` enabling similar local tween effects.


joshlory:
Worth extracting the `animationManager` change to a separate PR, and preflighting it to reduce the diff in this change?

src/js/game/Entities/Boat.js
```diff
@@ -30,5 +30,26 @@ module.exports = class Boat extends BaseEntity {
...
+   * which runs forever.
+   * @param {Phaser.Game} game
+   * @param {Phaser.Sprite} sprite
+   * @returns {Phaser.Tween}
+   */
+  static addBobTween(game, sprite) {
+    return game.add.tween(sprite).to(
+      { y: '-3' },
+      1000,
+      Phaser.Easing.Quadratic.InOut,
```
joshlory: Nit: sinusoidal?

islemaster: Are you saying you'd prefer `Phaser.Easing.Sinusoidal.InOut` to `Phaser.Easing.Quadratic.InOut`?  Any particular reason?  I'm cool with either one, the difference is pretty subtle.

joshlory: My thought was that sinusoidal might be closer to wave action, butI think either is fine!

joshlory:
Ah, I see now!  Also fine to merge as-is!

src/js/game/Entities/Boat.js
```diff
@@ -30,5 +30,26 @@ module.exports = class Boat extends BaseEntity {
...
+  /**
+   * Apply a "bob up and down in the water" animation to the sprite,
+   * which runs forever.
+   * @param {Phaser.Game} game
+   * @param {Phaser.Sprite} sprite
+   * @returns {Phaser.Tween}
+   */
+  static addBobTween(game, sprite) {
+    return game.add.tween(sprite).to(
+      { y: '-3' },
```
joshlory: Nit: should be a number?

PR #521, opened 2018-10-19
Minecraft | Aquatic mobs cannot be trampled in Freeplay levels

WilliamHarmonMS:


joshlory:
Closing this one out — Neal is ok with the current behavior on the free-play level, and I don't think we want to special-case this behavior unless we absolutely have to.

https://codedotorg.slack.com/archives/C6MJHRTGC/p1540497483000100

PR #522, opened 2018-10-19
Minecraft | Fixed the turtle animation

WilliamHarmonMS:
Bad Turtle:
badturtle

GOOD Turtle:
goodturtle

🍪 🐢


WilliamHarmonMS:
Offset was kinda crazy, which I assumed is because of the lack of trimming, but I'll take a weird offset with a smooth animation over something that's a bad user experience.

PR #523, opened 2018-10-19
Minecraft | OverOverlay will now scale correctly to 20x20

WilliamHarmonMS:
20x20 overlay:
20x20overlay

### DIFF
--- a/src/js/game/LevelMVC/LevelView.js
+++ b/src/js/game/LevelMVC/LevelView.js
@@ -632,16 +632,29 @@ module.exports = class LevelView {
     return animation;
   }
 
+  getOceanOverlayVertices(){
+    const gridDimensions = this.controller.levelData.gridDimensions;
+    if (gridDimensions[0] === 10 && gridDimensions[1] === 10) {
+      // This is a normal 10x10 map. Ocean Overlay will be tiled 4x
+      return [[0, 0], [200, 0], [0, 200], [200, 200]];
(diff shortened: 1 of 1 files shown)


src/js/game/LevelMVC/LevelView.js
```diff
@@ -632,16 +632,29 @@ module.exports = class LevelView {
...
   addOceanOverlay(tint) {
-    for (let [x, y] of [[0, 0], [200, 0], [0, 200], [200, 200]]) {
+    this.getOceanOverlayVertices()
+    for (let [x, y] of this.getOceanOverlayVertices()) {
       const surface = this.fluffGroup.create(x, y, "underwaterOverlay", "water_still_grey00.png");
       const frameList = Phaser.Animation.generateFrameNames("water_still_grey", 0, 31, ".png", 2);
       surface.animations.add("idle", frameList, 5, true);
       this.playScaledSpeed(surface.animations, "idle");
 
-      [surface.scale.x, surface.scale.y] = this.controller.scaleFromOriginal();
```
WilliamHarmonMS: the scale should always be [1,1] it seems, so the resulting surface scale comes out to [400 / 32, 400 / 32] every time. I simply hard coded this, because I didn't want to mess with scaleFromOriginal(). GitBlame says that was introduced years ago as a method involved in capturing screen shots so it seems like changing it might have unintended downstream effects for other systems.

PR #529, opened 2018-10-26
Underwater shader v2

joshlory:
Resurrect the underwater shader experiment from PR https://github.com/code-dot-org/craft/pull/474.

Right now it's essentially the same as the current shipped version, just a little more sparkle from properly implementing the "screen" blend mode.

There's room to add motion or distortion if we get approval from Joel.


joshlory:
image

joshlory:
Got the OK from Joel to try this out!

PR #533, opened 2018-10-30
Underwater update

joshlory:
Use the ocean "warm"/"cold" value instead of always being a fixed value.  Move the overlay relative to the camera on levels larger than 10x10.


test/helpers/AquaticLevels.js
```diff
@@ -173,7 +173,7 @@ module.exports = {
   },
   aquatic11: {
     isAquaticLevel: true,
-    ocean: "warm",
+    ocean: "cold",
```
islemaster: Unrelated - it's occurred to me that this file should probably be at `test/fixtures/AquaticLevels.js`

joshlory: It serves double-duty: we also use it for the levels in the playground!  https://code-dot-org.github.io/craft/?level=aquatic11

PR #536, opened 2018-10-30
Bubble column should be transparent for FoW calculation

joshlory:
Before:

After:


Hamms:
I'm not seeing the differences here; can you explain a bit?

joshlory:
Only the center area is lit before (from sea lanterns).  After, the magma also is lit.

joshlory:
Take a look at the lightness of the bottom left and bottom right corners!  (of the grey stone blocks, near the magma bubbles)

Hamms:
Ooooh, there it is! I had been looking at the floor-level blocks more than at the play-level blocks

PR #540, opened 2018-11-01
Allow two-corner illumination

### LINKED ISSUE #537: FoW edge case: no tile with two opposite corners lit
image

We don't account for two opposite corners being lit, with the remaining two corners in darkness:

https://github.com/code-dot-org/craft/blob/5a8b2378b4bfb0e7d9d68de4b768ca947455e9df/src/js/game/LevelMVC/LevelModel.js#L902-L918

islemaster:
Fixes #537

Adds two new textures to the fog-of-war atlas for "diagonal" illumination, tiles illuminated from opposite corners.

| Before | After |
| --- | --- |
| screenshot from 2018-10-31 22-14-11 | screenshot from 2018-10-31 22-21-22 |

Just now realizing that this may not solve the issue as described by Josh's additional comment on that issue, but it does look better.


src/js/game/LevelMVC/LevelView.js
```diff
@@ -1858,14 +1858,6 @@ module.exports = class LevelView {
...
         fy = -40 + 40 * fowItem.y;
 
-        switch (fowItem.type) {
-          case "FogOfWar_Center":
-            break;
-
-          default:
-            break;
-        }
-
```
islemaster: Unrelated dead code.

islemaster:
[Minecraft: Allow two-corner illumination](https://trello.com/c/LzQ1Sywo/27-minecraft-allow-two-corner-illumination)

joshlory:
> Just now realizing that this may not solve the issue as described by Josh's additional comment on that issue, but it does look better.

Nope, this was exactly what I was thinking!  We’ve had the current FoW calculation since 2015, it’s too risky to change it this late in the game.  This looks 1000x better, with minimal risk!

PR #551, opened 2019-03-05
Allow WASD keys and Spacebar to be used in project names

ajpal:


src/js/game/GameController.js
```diff
@@ -277,18 +278,30 @@ class GameController {
       [Phaser.Keyboard.SPACEBAR]: -2
     };
 
+    const shouldIgnoreKeyPress = function () {
+      // Ignore keyboard input when project is being renamed
+      return $('.project_save').length > 0;
```
Hamms: I don't love the idea of polluting the craft repo with specific knowledge about the containing system. What if we did something like

```js
function editableElementSelected () {
  return !(document.activeElement === document.body || document.activeElement === null);
}
```

src/js/game/GameController.js
```diff
@@ -277,18 +277,29 @@ class GameController {
       [Phaser.Keyboard.SPACEBAR]: -2
     };
 
+    const editableElementSelected = function () {
+      return !(document.activeElement === document.body || document.activeElement === null);
```
joshlory: Hmm I'm a little worried this test might be too general. If I press "tab" on https://studio.code.org/s/minecraft/stage/1/puzzle/12, I get:

image

which isn't `null` or `document.body`.

